PR #3, opened 2021-12-21
Test consistency of static GTFS feeds for live agencies

unreasonableman:
`LoadAgencyDataTest` pulls a list of agencies from a configurable URL and checks the consistency of their static GTFS data. Failures are recorded and once the test completes, these failures are emailed to recipients listed in the environment variable `GRAAS_REPORT_RECIPIENTS`. 

No slack integration at the moment, since our free plan is constantly out of slack apps, of which we would need one :[

### DIFF
--- a/gtfu/src/test/java/gtfu/test/LoadAgencyDataTest.java
+++ b/gtfu/src/test/java/gtfu/test/LoadAgencyDataTest.java
@@ -27,7 +27,7 @@ public class LoadAgencyDataTest {
         ConsoleProgressObserver progressObserver = new ConsoleProgressObserver(40);
         Util.updateCacheIfNeeded(cacheDir, agencyID, gtfsUrl, progressObserver);
 
-        Map<String, Object> collections = Util.loadCollections(cacheDir, agencyID, progressObserver);
+        Map<String, Object> collections = Util.loadCollections(cacheDir, agencyID, progressObserver, true);
 
         TripCollection tripCollection = (TripCollection)collections.get("trips");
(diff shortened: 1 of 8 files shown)


gtfu/src/test/java/gtfu/test/LoadAgencyDataTest.java
```diff
@@ -38,12 +41,27 @@ public LoadAgencyDataTest(String cacheDir, String agencyID) throws Exception {
 
     private static void usage() {
         System.err.println("usage: LoadAgencyDataTest -c|--cache-dir <cache-dir> -a|--agency-id <agency-id>");
+        System.err.println("usage: LoadAgencyDataTest -c|--cache-dir <cache-dir> -u|--url <url>");
+        System.err.println("    <url> is assumed to point to a plain text document that has an agency ID per line");
```
owades, line 45: Do we have a way to automatically generate the URL/text file, or is this another agency onboarding step?

owades (approved):
Left one question about manual work required to generate text file

gtfu/src/test/java/gtfu/test/LoadAgencyDataTest.java
unreasonableman: I think for now we want to manually curate it, so that new agencies are added but also so that dusty agencies get removed. Feel free to create something clever in the future that automates this, but for now I think the approach should be: when onboarding a new agency, copy file from bucket, add agency ID and look for old agencies to remove, then copy back to bucket

PR #8, opened 2021-12-23
Launch daily error reporting for static feeds

owades:
1. Update LoadAgencyDataTest to check for null ShapeIDs
2. Run LoadAgencyDataTest daily via GH actions
3. Create new Recipients class, to prevent duplicative work & textfiles. Update LoadAgencyDataTest and GraphicReport to use Recipients class

Notes:
- We can host the recipients.txt in a URL obfuscated via GH secrets if we so choose (this is how we handle the active-agencies file).
- I haven't tested the LoadAgencyDataTest on an agency with null ShapeID, so I'd appreciate either A) insistence that I do so or B) a thorough look at that code
- We might consider merging this with GRaaS Report M-F GH Action, but that would require running 5 days a week versus daily.


unreasonableman:
Notes on your notes:
- untested code is buggy code. You can pretty easily test the shape ID stuff by manually editing a trips.txt file in your cache folder and just changing a few shape IDs to not match anymore. Confirm that everything works as expected, including the test's ability to find multiple issues per agency
- regarding `recipients.txt`: currently the entire bucket is publicly readable. It would be great to a) make it so that there are parts of the bucket that are private, and consequently have a way for us to authenticate for access to those parts

gtfu/src/main/java/gtfu/Recipients.java
```diff
@@ -0,0 +1,33 @@
...
+	public Recipients() {
+
+        recipientsMap = Util.hashMapFromTextFile(RECIPIENTS_FILE_PATH,DELIMITER);
+	}
+
+    public String[] get(String reportType) {
+
+        String recipientsList = null;
+        // Default to local recipient list if env variable is present
+        recipientsList = System.getenv("LOCAL_REPORT_RECIPIENTS");
```
unreasonableman, line 20: There isn't really a point to initializing to `null` if you're reassigning in the very next step. `System.getenv()` returns `null` for unset environment variables.

gtfu/src/main/java/gtfu/Recipients.java
```diff
@@ -0,0 +1,33 @@
...
+
+        if (recipientsList == null) {
+        	recipientsList = recipientsMap.get(reportType);
+
+			if (recipientsList == null) {
+				System.out.println("** error: no report type " + reportType + " listed in " + RECIPIENTS_FILE_PATH);
+				System.exit(1);
```
unreasonableman, line 27: Calling `System.exit()` is a very bad idea for code that isn't a top-level application. You don't know who might be calling this in the future, and there is no stack trace to explain what happened. Just throw an exception.

gtfu/src/main/java/gtfu/Util.java
```diff
@@ -846,4 +846,32 @@ public static String getGTFSURL(AgencyData[] list, String id) {
         return collections;
     }
 
-}
\ No newline at end of file
+    public static Map<String, String> hashMapFromTextFile(String path, String delimiter) {
+
+        Map<String, String> map = new HashMap<String, String>();
```
unreasonableman, line 851: Nitpick: you can just say `new HashMap()`

gtfu/src/main/java/gtfu/Util.java
```diff
@@ -846,4 +846,32 @@ public static String getGTFSURL(AgencyData[] list, String id) {
...
+                String recipients = parts[1].trim();
+
+                if (!name.equals("") && !recipients.equals("")){
+                    map.put(name, recipients);
+                }
+            }
+        }
+        catch (Exception e) {
+            System.out.println("** error: no such file " + path);
+            System.exit(1);
```
unreasonableman, line 872: See my other comment about calling `exit()` from non-application-level code

gtfu/src/main/java/gtfu/Util.java
```diff
@@ -846,4 +846,32 @@ public static String getGTFSURL(AgencyData[] list, String id) {
...
+        try {
+            File file = new File(path);
+            br = new BufferedReader(new FileReader(file));
+            String line = null;
+
+            while ((line = br.readLine()) != null) {
+
+                String[] parts = line.split(delimiter);
+                String name = parts[0].trim();
+                String recipients = parts[1].trim();
```
unreasonableman, line 863: This code purports to be a general routine, but these names assume a very specific purpose. Instead of `name` and `recipients`, use something like `key` and `value`

gtfu/src/main/java/gtfu/Util.java
```diff
@@ -846,4 +846,32 @@ public static String getGTFSURL(AgencyData[] list, String id) {
...
+            }
+        }
+        catch (Exception e) {
+            System.out.println("** error: no such file " + path);
+            System.exit(1);
+        }
+
+        return map;
+    }
+}
```
unreasonableman, line 877: I don't like this approach. It's very limited. Instead, please consider:

- recipients.json:
```
    {
      'error_report': [
        'scott.owades@dot.ca.gov', 
        'kay.neuenhofen@dot.ca.gov'
      ],
      'graas_report': [
        'scott.owades@dot.ca.gov', 
        'lauren.gilbert@rebelgroup.com, 
        'ali.attari@rebelgroup.com', 
        'kay.neuenhofen@dot.ca.gov'
      ]
  }
```

- Util.java:
  ```
  public static Map<String, Object> parseJSONasMap(String path) {
        try {
            return new ObjectMapper().readValue(path, HashMap.class);;
        } catch (IOException e) {
            throw new Fail(e);
        }
  }
```

**Review by unreasonableman — CHANGES_REQUESTED** [review]
Please see comments

File: gtfu/src/main/java/gtfu/Util.java (hunk 5 above)
Comment (owades) [reply gtfu/src/main/java/gtfu/Util.java:863]: Crap. I went back and forth on whether to include in Util, forgot to change the names back :-(

**Comment by owades** [comment]
Thanks again for this feedback, I completed the requested items. One thing I did not do yet is create any kind of authentication/obfuscation for recipients.json.

File: gtfu/src/main/java/gtfu/Util.java [hunk 7]
```diff
@@ -846,4 +847,14 @@ public static String getGTFSURL(AgencyData[] list, String id) {
         return collections;
     }
 
-}
\ No newline at end of file
+    public static Map<String, Object> parseJSONasMap(String path) {
+        try {
+            TypeReference<HashMap<String, String[]>> typeRef
+              = new TypeReference<HashMap<String, String[]>>() {};
```
unreasonableman, line 853: This would mean we're limiting ourselves to parsing JSON objects that contain lists of strings as their values, but we want to be able to parse all kinds of objects, including nested ones.

gtfu/src/main/java/gtfu/Recipients.java
```diff
@@ -0,0 +1,33 @@
...
+
+	private static final String RECIPIENTS_FILE_PATH = "src/main/resources/conf/recipients.json";
+	private Map<String, Object> recipientsMap;
+
+	public Recipients() {
+        recipientsMap = Util.parseJSONasMap(RECIPIENTS_FILE_PATH);
+	}
+
+    public String[] get(String reportType) {
+
```
unreasonableman, line 15: The indentation and whitespace for much of this method looks off. 

Maybe we should do the pre-commit hook formatting after all that E was talking about, to avoid this tediousness.

unreasonableman (changes requested):
Please see my comment on `parseJSONasMap`

gtfu/src/main/java/gtfu/Util.java
owades: This bit turned out to be the only tough part of the edits: getting the object converted to an array of strings without making things too ugly. This is the cleanest I could get it, but I can go back to the drawing board based on this feedback.

unreasonableman: It seems much more elegant to be able to parse arbitrary JSON objects into a map. I would assume the array of recipients in the JSON is parsed into a java Collection object. If that's the case, you should be able to use `Collection.toArray()`. If for some reason that's not possible then add a little utility function to `Recipients` that does the conversion.

unreasonableman: Please do the right thing and then merge. Unless you want to have this PR sitting until Jan 3...

unreasonableman (approved):
Not sure you can merge as long as the status is `changes request`, so changing, please see my last comments.

gtfu/src/main/java/gtfu/Recipients.java
owades: Updated indentation and created https://github.com/cal-itp/graas/issues/27 for automation

owades:
@unreasonableman , ready for hopefully one last review. Util.parseJSONasMap() creates a map of Strings to Objects, and Recipients.java converts the map to an Array[]. It does so via a for loop, which I don't love, but since it needs to convert each ArrayList item from Object to String I believe this is necessary.

gtfu/src/main/java/gtfu/Recipients.java
```diff
@@ -0,0 +1,44 @@
...
+				String[] recipientsList = new String[recipients.size()];
+
+				for (int i = 0; i < recipients.size(); i++) {
+					recipientsList[i] = (String) recipients.get(i);
+				}
+
+				return recipientsList;
+
+			} catch (Exception e) {
+	            throw new Fail(e);
```
unreasonableman, line 40: We need to figure out the indent situation. Let's talk about it in our meeting tomorrow.

gtfu/src/main/java/gtfu/Recipients.java
```diff
@@ -0,0 +1,44 @@
...
+				throw new Fail("** error: no report type " + reportType + " listed in " + RECIPIENTS_FILE_PATH);
+			}
+
+			// Convert arraylist of objects to array of strings
+			try{
+				String[] recipientsList = new String[recipients.size()];
+
+				for (int i = 0; i < recipients.size(); i++) {
+					recipientsList[i] = (String) recipients.get(i);
+				}
```
unreasonableman, line 35: recipients.toArray() would do the same thing

gtfu/src/main/java/gtfu/TripCollection.java
```diff
@@ -35,10 +40,21 @@ public TripCollection(String path, StopCollection stopCollection, ShapeCollectio
             String serviceID = r.get("service_id");
             String headSign = r.get("trip_headsign");
             String shapeID = r.get("shape_id");
-            Trip trip = new Trip(id, routeID, serviceID, headSign, shapeCollection.get(shapeID));
-
-            map.put(id, trip);
-            list.add(trip);
+            Shape shape = shapeCollection.get(shapeID);
+            if (shape == null) {
```
unreasonableman, line 44: Is this branch tested?

unreasonableman (approved):
Assuming TripCollection changes are tested.

gtfu/src/main/java/gtfu/TripCollection.java
owades: Yep - tested thoroughly and successfully caught multiple missing shapes across multiple agencies

gtfu/src/main/java/gtfu/Recipients.java
owades: replied to slack message about this - i wasn't able to get it to work because the arraylist items were objects, and needed to be casted to string.

unreasonableman:
👍

PR #24, opened 2021-12-23
remove static agency data file and update file that relied on it

owades:
We're no longer using or updating agency-data.json, so I thought I'd throw it out.

I'm not familiar with RealtimeVisualization.java, so for now I just ran usage() instead of trying to load static data. If this is something that inconveniences your workflow we could leave it as-is, or update to use AgencyYML


gtfu/src/main/java/gtfu/RealtimeVisualization.java
```diff
@@ -752,12 +752,7 @@ public static void main(String[] arg) {
         }
 
         if (list == null) {
-            try {
-                InputStream is = RealtimeVisualization.class.getResourceAsStream("/conf/agency-data.json");
-                list = (AgencyData[])Util.readJSONObjectFromStream(is, AgencyData[].class);
-            } catch (Exception e) {
-                throw new Fail(e);
-            }
+            usage();
```
unreasonableman, line 755: I think your idea of using AgencyYML is very interesting. You would have to iterate over the returned data and filter for agencies that have both a static and an RT feed, but it would a) expand the scope of agencies we can visualize traffic for, and b) it would be more up-to-date than a static file. If you don't feel like doing that now, please create an issue for it.

unreasonableman (approved):

gtfu/src/main/java/gtfu/RealtimeVisualization.java
owades: Makes sense -  it would require adding GRaaS RT URL to agencies.yml, which is something we'd need to talk to with the file owners.

Created https://github.com/cal-itp/graas/issues/25

unreasonableman: It's already in there. Excerpt for CAE:

```
agency_name: Clean Air Express
  feeds:
    - gtfs_schedule_url: http://www.cleanairexpress.com/GTFS/GTFS.zip
      gtfs_rt_vehicle_positions_url: null
```

PR #37, opened 2022-01-05
Set up unit testing in Java & Python

owades:
- Set up Java unit testing via jUnit, with two simple tests
- Set up Python unit testing with unittest, with two simple tests
- Incorporate Java testing into default gradle build (which runs on each PR)
- Incorporate python testing to run on each PR


gtfu/src/test/java/gtfu/test/TestAgencyName.java
```diff
@@ -0,0 +1,22 @@
...
+class TestAgencyName {
+
+    @Test
+    @DisplayName("TestAgencyName")
+    void testName() {
+        AgencyYML agencies = new AgencyYML();
+        // Not ideal to hardcode an agency name into unit test, but it also
+        // wouldn't be a good idea to add test agencies into agencies.yml.
+        // Open to better solutions here.
+        assertEquals("Tulare County Regional Transit Agency", agencies.getName("tcrta"));
```
unreasonableman, line 20: It is probably marginally better to use a bigger agency here (e.g. BART), both because they already have exposure and they are more stable

unreasonableman (approved):
The trick is to keep adding tests going forward.

gtfu/src/test/java/gtfu/test/TestAgencyName.java
owades: that's a good idea, thanks. done.

PR #44, opened 2022-01-10
Docs/fix readme urls

owades:
I noticed that URL in readme was working in MkDocs but not via github directly. Confirmed that fix works when viewing from github or mkdocs.


unreasonableman (approved):
Nice!

PR #45, opened 2022-01-11
switch to daily not just weekdays

owades:


unreasonableman (approved):
I don't know the first thing about action yaml files. The changes are in lines that look like competing ways to describe a schedule. Is the top one just for display? Even then, it seems much better to automatically generate from the 'cron' line.

owades:
The top line is just a comment, to translate cron to english and to PST. The rest (on, schedule, cron) is standard GH action syntax

PR #47, opened 2022-01-11
Features/trip inference

unreasonableman:
merge latest changes for trip inference. Still WIP.


gtfu/src/main/java/gtfu/tools/TripInferenceVisualizer.java
```diff
@@ -404,11 +426,93 @@ public void latLongToScreenXY(int displayWidth, int displayHeight, ShapePoint p)
         //Debug.log("- p.screenY: " + p.screenY);
     }
 
+    public void screenXYToLatLong(int displayWidth, int displayHeight, ShapePoint p) {
```
owades, line 429: It's a little odd that we have a latLongToScreenXY() function in GraphicReport.java, which works differently (returns a value rather than updating a value). Could both functions live in the same place and/or work similarly.

gtfu/src/main/java/gtfu/tools/TripInferenceVisualizer.java
```diff
@@ -404,11 +426,93 @@ public void latLongToScreenXY(int displayWidth, int displayHeight, ShapePoint p)
...
+        float fractionX = (float)p.screenX / displayWidth;
+        float fractionY = (float)p.screenY / displayHeight;
+        Area area = areaStack.peek();
+
+        p.lat = area.topLeft.lat - area.getLatDelta() * fractionY;
+        p.lon = area.topLeft.lon + area.getLongDelta() * fractionX;
+    }
+
+    public void mouseClicked(MouseEvent e) {
```
owades, line 438: TODO?

owades (approved):
I flagged two things, but guessing that you have answers to both.  Curious to hear.

gtfu/src/main/java/gtfu/tools/TripInferenceVisualizer.java
unreasonableman: There are some nuances around aspect ratio that I've never been able to work out in a one-size-fits-all fashion. I agree that ideally the method should just exist once.

gtfu/src/main/java/gtfu/tools/TripInferenceVisualizer.java
unreasonableman: What's the TODO here?

PR #50, opened 2022-01-13
add new attributes to graphicreport

owades:
- Rename "name" to "headsign" in cases where it actually is the headsign. I have flip-flopped on this in past PRs but believe we're moving towards consistency
- Add RouteCollection is parameter to DayLogSlicer, and have it provide route_name when trip_headsign is null. This required updating TrainingDataUtil.java, something I'm not familiar with and don't know how to test. But ultimately it just replaces a string value with another string value
- Use [ua_parser](https://github.com/ua-parser/uap-java) to parse useragent, os & device from agent field
- Add agent, os, device, uuid & vehicleID to GRaaS report

Potential issues 
- Currently user agent parsing is occurring within TripReportData.java. There is a good argument for creating a file specifically for parsing user agents. That work could be done in this PR, or once we have a second use case for these attributes. Left a TODO comment with this info.
- New attributes listed on graphicreport (device, vehicleid, uuid tail, os, agent) could potentially overflow beyond border of chart. This PR doesn't have dynamic capabilities for adjusting or resizing text. Left a TODO comment with this info.

Example report:
Screen Shot 2022-01-13 at 1 14 16 PM

### DIFF
--- a/gtfu/src/main/java/gtfu/GraphicReport.java
+++ b/gtfu/src/main/java/gtfu/GraphicReport.java
@@ -43,6 +43,7 @@ public class GraphicReport {
     private static final Color BACKGROUND = new Color(0xffffff);
     private static final Color DARK       = new Color(0xe0e0e0);
     private static final Color FONT_COLOR = Color.gray;
+    private static final Color TITLE_COLOR = Color.black;
     private static final Color ACCENT     = new Color(0x00b000);
     /*private static final Color BACKGROUND = new Color(0x1c3a08);
     private static final Color DARK       = new Color(0x324e34);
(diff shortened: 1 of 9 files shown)


unreasonableman:
> could potentially overflow beyond border of chart.

As a first workaround, please use `Graphics.setClip()` to prevent tile spill. Be sure to reset clip to entire image once done with each tile. 

I'll review once that's done.

owades:
Added clip. Tested with some filler text, successfully clipped it. And I even remembered to take out the filler text 
Screen Shot 2022-01-13 at 3 23 15 PM

gtfu/src/main/java/gtfu/GraphicReport.java
```diff
@@ -253,7 +257,7 @@ private void addRandomTestData(float prob) {
                     int t2 = t.getTimeAt(t.getStopSize() - 1);
                     int duration = t2 - t1;
 
-                    TripReportData td = new TripReportData(id, t.getName(), start, duration);
+                    TripReportData td = new TripReportData(id, t.getHeadsign(), start, duration,"testUuid","testAgent","testVehicleId");
```
unreasonableman, line 260: Let's use consistent white-spacing, e.g. "a, b" and not "a,b"

unreasonableman (approved):

PR #51, opened 2022-01-14
dotsize reflects repeat gps updates

owades:
- GPS update dot size increases by 20% of original size for each repeat update sharing same exact latlong. Did this by replacing List<GPSData> with Map<String,GPSData>, which maps each latLong combo with the corresponding GPSData. Increment when already exists.
- Fixes bug where DesertRoadrunner trip names weren't appearing

Potential Issues:
- TrainingDataUtil.java changes remain untested. Will seek guidance on how to test.

### DIFF
--- a/gtfu/src/main/java/gtfu/GPSData.java
+++ b/gtfu/src/main/java/gtfu/GPSData.java
@@ -4,11 +4,13 @@ public class GPSData {
     public long millis;
     public float lat;
     public float lon;
+    public int count;
 
-    public GPSData(long millis, float lat, float lon) {
+    public GPSData(long millis, float lat, float lon, int count) {
         this.millis = millis;
         this.lat = lat;
         this.lon = lon;
+        this.count = count;
     }
 
     public String toString() {
@@ -17,5 +19,9 @@ public class GPSData {
 
     public String toCSVLine() {
         return String.format("%d,%f,%f", millis / 1000, lat, lon);
+    }
+
+    public void increment() {
+        count++;
     }
 }
--- a/gtfu/src/main/java/gtfu/GPSData.java
+++ b/gtfu/src/main/java/gtfu/GPSData.java
@@ -6,11 +6,11 @@ public class GPSData {
     public float lon;
     public int count;
 
-    public GPSData(long millis, float lat, float lon, int count) {
+    public GPSData(long millis, float lat, float lon) {
         this.millis = millis;
         this.lat = lat;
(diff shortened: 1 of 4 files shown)


gtfu/src/main/java/gtfu/GPSData.java
```diff
@@ -4,11 +4,13 @@
     public long millis;
     public float lat;
     public float lon;
+    public int count;
 
-    public GPSData(long millis, float lat, float lon) {
+    public GPSData(long millis, float lat, float lon, int count) {
```
owades, line 9: It would be cleaner to set it to 1 and not require count as a parameter.

unreasonableman: Don't know if I'm understanding correctly, but you can overload constructors:
```
public GPSData(long millis, float lat, float lon) {
    this(long millis, float lat, float lon, 1);
}

public GPSData(long millis, float lat, float lon, int count) {
    ...
}
```

gtfu/src/main/java/gtfu/tools/DayLogSlicer.java
```diff
@@ -56,14 +57,21 @@ public DayLogSlicer(TripCollection tripCollection, RouteCollection routeCollecti
...
-                gpsMap.put(tripID, list);
+            String latLon = String.valueOf(lat) + String.valueOf(lon);
+            if(gpsMap.get(tripID) == null){
+                Map<String, GPSData> latLongMap = new HashMap();
+                latLongMap.put(latLon, new GPSData(seconds * 1000l, lat, lon, 1));
+                gpsMap.put(tripID,latLongMap);
+            }
+            else {
+                if (gpsMap.get(tripID).get(latLon) == null) {
+                    gpsMap.get(tripID).put(latLon, new GPSData(seconds * 1000l, lat, lon, 1));
```
unreasonableman, line 69: I don't understand why there are these two very similar if statements here, please add a comment.

unreasonableman (changes requested):
Looks good mostly, add request for explanation for one bit.

owades:
@unreasonableman ready for re-review. You were right about the redundant logic - I cleaned it up, and also added comments.

unreasonableman (approved):

owades:
@unreasonableman would you prefer I hold on merging until we have addressed needed changes to TrainingDataUtil?

unreasonableman:
Go ahead and merge.

PR #55, opened 2022-01-15
Create unit test to confirm LoadAgencyDataTest is working as expected

owades:
- Fix bug in LoadAgencyDataTest, by updating TripScheduleCollection and TripCollection
- Update FailureReporter to show count of fails
- Create new unit test, which confirms the number of fails is equal to the manually-set expected number
- Add a cache directory with "test-agency" static gtfs files


gtfu/src/main/java/gtfu/TripCollection.java
```diff
@@ -80,6 +80,9 @@ public TripCollection(String path, StopCollection stopCollection, ShapeCollectio
                 if (Util.isEmpty(time)) continue;
 
                 trip = map.get(tripID);
+
+                if (trip == null) continue;
```
owades, line 84: Fail message was sent in above lines, this "continue" is so we can work with the trips that DON'T lack a shape

owades: Note that this bug prevented missing-route errors for showing in addition to missing-shape errors.

unreasonableman (changes requested):
- why is agency.txt empty? Unless there's a specific reason, it'd be good to have data in there
- I don't really understand your comments around the added `continue` in TripCollection. Maybe we can talk about that in the triage meeting
- You have `gtfu/src/test/resources/conf/cache/test-agency` as the resource path for the bad static gtfs test. I think `gtfu/src/test/resources/bad-static-gtfs` might make more sense, as a) it's easier to correlate data and code and b) I'm not sure what `conf/cache` in the path really adds

PR #56, opened 2022-01-19
add gps update stats and dot opacity to GRaaS report

owades:
- Added functionality to calculate and present basic stats on gps updates: avg time between updates, min & max update time.
- As dot size increases, now transparency does as well. This makes larger dots less disruptive.
- Confirmed that GraphicReport.addRandomTestData() still works


gtfu/src/main/java/gtfu/tools/DayLogSlicer.java
```diff
@@ -59,15 +63,21 @@ public DayLogSlicer(TripCollection tripCollection, RouteCollection routeCollecti
...
                 gpsMap.put(tripID,latLonMap);
             }
-            // If there is no existing GPSData for this latLon value, add it
+
+            // If there is no existing GPSData for this latLon value, add it and update previousUpdate for that trip
             if (gpsMap.get(tripID).get(latLon) == null) {
-                gpsMap.get(tripID).put(latLon, new GPSData(seconds * 1000l, lat, lon));
+                // previousUpdateMap stores the most recent GPS update timestamp (in seconds) for each trip_id.
+                // it relies on the list being sorted by timestmap, which it is.
+                gpsMap.get(tripID).put(latLon, new GPSData(seconds * 1000l, ((previousUpdateMap.get(tripID) != null) ? seconds - previousUpdateMap.get(tripID) : -1), lat, lon));
```
owades, line 77: I could see an argument for being too complex of a line - can calculate this previous update time in a previous line.

unreasonableman:
> As dot size increases, now opacity does as well. This makes larger dots less disruptive.

Opacity decreases. Transparency decreases </pedantery>

gtfu/src/main/java/gtfu/GPSStats.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public GPSStats(Map<String, GPSData> latLonMap) {
+        int runningTotalUpdates = 0;
+        int runningTotalTimeSinceUpdate = 0;
+        int secsSinceLastUpdate = 0;
+
+        for (String latLon : latLonMap.keySet()) {
+            secsSinceLastUpdate = latLonMap.get(latLon).secsSinceLastUpdate;
+            if(secsSinceLastUpdate >= 0) {
+                runningTotalUpdates += 1;
```
unreasonableman, line 24: You can just say `runningTotalUpdates ++`.

gtfu/src/main/java/gtfu/GPSStats.java
```diff
@@ -0,0 +1,48 @@
...
+import java.text.DecimalFormat;
+
+public class GPSStats {
+
+    private static final DecimalFormat df = new DecimalFormat("0.00");
+    public int maxUpdateTime = 0;
+    public int minUpdateTime = 999999;
+    public float avgUpdateTime = 0;
+
+    public GPSStats(Map<String, GPSData> latLonMap) {
```
unreasonableman, line 16: It would be good form to have a comment that says this constructor depends on the lat long map being ordered by time

gtfu/src/main/java/gtfu/GPSStats.java
```diff
@@ -0,0 +1,48 @@
+package gtfu;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
+
+public class GPSStats {
```
unreasonableman, line 9: Nothing about this is specific to GPS. Can it just be a stats class?

gtfu/src/main/java/gtfu/GraphicReport.java
```diff
@@ -489,11 +497,16 @@ private void drawMap(Graphics2D g, TripReportData td, int length) {
 
             Integer count = latLonMap.get(latLon).count;
             float scaledDotSize = DOT_SIZE * (1 + (count - 1) / DOT_SIZE_MULTIPLIER);
+            double alpha = Math.pow(OPACITY_MULTIPLIER, (double) (count - 1) );
+            float alphaFinal = (float) ((alpha >= ALPHA_MIN) ? alpha : ALPHA_MIN);
```
unreasonableman, line 501: nitpick: `alpha = Math.max(ALPHA_MIN, alpha)` is more concise.

gtfu/src/main/java/gtfu/tools/DayLogSlicer.java
unreasonableman: Yeah, that's pretty ugly.

unreasonableman (changes requested):
Please see comments

gtfu/src/main/java/gtfu/GPSStats.java
owades: It doesn't, though! DayLogSlicer does rely on the query from DB.java being ordered by timestamp. I mention this in a comment in DayLogSlicer, but I'll add a comment in DB.java as well.

gtfu/src/main/java/gtfu/GPSStats.java
owades: Yes - though I'm not sure how to accept a generic object class as input rather than GPSData. Updating to Stats for now.

gtfu/src/main/java/gtfu/GraphicReport.java
owades: Wow, yeah. Thanks.

gtfu/src/main/java/gtfu/Stats.java
```diff
@@ -0,0 +1,51 @@
...
+            if (value < min) {
+                min = value;
+            }
+
+            if (value > max) {
+                max = value;
+            }
+        }
+
+        avg = total / count;
```
unreasonableman, line 33: Finding a bug in my own code. This should be:
`avg = count == 0 ? 0 : total / count;`
to avoid an exception if the collection is empty or all calls to `getValue()` return null

unreasonableman: Actually, avg gets initialized to `0` by the VM, so to make things a little less cryptic, you can just say:

`if (count > 0) avg = total / count;`

gtfu/src/main/java/gtfu/tools/DayLogSlicer.java
```diff
@@ -59,15 +63,25 @@ public DayLogSlicer(TripCollection tripCollection, RouteCollection routeCollecti
...
+            // If there is no existing GPSData for this latLon value, add it and update previousUpdate for that trip
             if (gpsMap.get(tripID).get(latLon) == null) {
-                gpsMap.get(tripID).put(latLon, new GPSData(seconds * 1000l, lat, lon));
+                // previousUpdateMap stores the most recent GPS update timestamp (in seconds) for each trip_id.
+                // it relies on the list being sorted by timestmap, which it is.
+                int secsSinceLastUpdate = -1;
+                if (previousUpdateMap.get(tripID) != null) {
+                    secsSinceLastUpdate = seconds - previousUpdateMap.get(tripID);
+                }
+                gpsMap.get(tripID).put(latLon, new GPSData(seconds * 1000l, secsSinceLastUpdate, lat, lon));
```
unreasonableman, line 81: The expression  `gpsMap.get(tripID)` appears at least four times in the code above. Most modern compilers have Common Subexpression Elimination, so likely not a performance issue, but still overly verbose and a potential cause of errors. Just assign to a variable up top and then reference that

unreasonableman (changes requested):
Please fix divide by zero error pointed out in comment, and the repeated expression in the second comment, after that you're good to go

PR #57, opened 2022-01-19
Having the baby shark client code in the main repo turns out to be a …

unreasonableman:
…bad idea. Sharks are often on unrealiable, limited-bandwidth network, which makes cloning and pulling from even mid-size repos arduous. As a remedy, client code is split out into https://github.com/cal-itp/graas-hardware.git.

Updated docs and moved STL files needed for case printing.


owades:
@unreasonableman  have you tried cloning just the subdirectory? I searched briefly and saw [this recent-ish article](https://terminalroot.com/how-to-clone-only-a-subdirectory-with-git-or-svn/) about doing so.

unreasonableman:
@owades That would still have the STL files, which are kind of chunky. 

I appreciate the tip, but:
- the approach described in the article seems convoluted
- I'm not sure you wouldn't still pull the git history for the entire repo
- the article doesn't seem very polished which to me raises some questions as to how thorough the testing was
- the BBS client code is actually a totally separate thing and it makes sense to me to have it in a different repo

owades (approved):
Thanks for talking that through. Just wanted to make sure you looked down that route since one repo might be more elegant.

Don't forget to deploy mkdocs if you haven't already!

PR #59, opened 2022-01-21
- custom logic for BBS user agents

unreasonableman:
- test for said logic


owades (approved):
Nice! Thanks for doing this.

PR #60, opened 2022-01-24
- don't add updates with invalid GPS values to DB or vehicle position…

unreasonableman:
… streams

- fix missing class name for method call


owades (approved):
Looks good.

PR #62, opened 2022-01-25
add DOW as utm param, replace naked strings

owades:
- Added DOW as a UTM parameter
- Replaced naked strings that appear more than once with symbolic constants.
- There are one or two dozen more one-off strings that could be replaced as well

### DIFF
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -53,6 +53,14 @@ const CONFIG_ROUTE_NAMES = "route names";
 const CONFIG_VEHICLE_IDS = "vehicle IDs";
 const CONFIG_DRIVER_NAMES = "driver names";
 const CONFIG_FILTER_PARAMS = "filter params";
+const START_STOP_BUTTON = "start-stop";
+const START_STOP_BUTTON_LOAD_TEXT = "Load trips"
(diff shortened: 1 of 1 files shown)


server/app-engine/static/graas.js
```diff
@@ -348,11 +360,11 @@ function handleRefreshOkay() {
 function handleStartStop() {
     util.log("handleStartStop()");
 
-    var p = document.getElementById('start-stop');
+    var p = document.getElementById(START_STOP_BUTTON);
     var text = p.textContent || p.innerText;
     util.log("- text: " + text);
 
-    if (text == "Load trips") {
+    if (text == START_STOP_BUTTON_LOAD_TEXT) {
```
unreasonableman, line 367: js has `==` behavior that is unintuitive to me where it does type conversion. `===` is almost always what we want instead of the former.

unreasonableman (changes requested):
Please see comment

server/app-engine/static/graas.js
owades: Thank you for catching this - clearly ==, ===, equals() etc  is an area I should focus on more, will do so. Updated, tested, will merge now.

PR #63, opened 2022-01-25
json input for stresstest params

owades:
The only functional difference is that stress test now defaults to local server, unless config file says production-server = true


server/app-engine/stress-test.py
```diff
@@ -133,41 +136,34 @@ def main(argv):
...
 		exit(1)
 
+	f = open(jsonInput)
+	data = json.load(f)
+	f.close()
+
+	if data.get('production-server', False):
+		domain = PRODUCTION_SERVER_URL
+	else:
+		domain = LOCAL_SERVER_URL
```
unreasonableman, line 158: Just a thought, but wouldn't it be more flexible if you just had a `server_url` attribute in the json? If not set, it defaults to `localhost`. That way, if we ever have a real staging env we can easily point to that, and also other people using the repo can point to whatever servers they have set up.

unreasonableman (approved):

server/app-engine/stress-test.py
owades: That's a nice idea and will add, thanks

owades:
I added an additional field to json input called **production-server-url**, and also renamed the boolean to **use-production-server**.

The reason I'm keeping them as separate fields is it's easier to toggle production on and off, rather than needing to delete the JSON attribute and re-add it.

PR #76, opened 2022-01-31
add session_id reporting to web app

owades:
It turns out no server-side changes are required in order to begin sending session-id to the DB (gtfsrt-util.py passes the whole **data** object to the DB). This change creates session-id with each time "okay" is pressed on app, and passes it to the server. Also updates get-pos.py for pulling session-id.


server/app-engine/get-pos.py
```diff
@@ -36,10 +36,11 @@ def main(argv):
     print_trip_id = False
     print_driver = False
     print_vehicle_id = False
+    print_session_id = False
     print_id = False
 
     try:
-        opts, args = getopt.getopt(argv,"u:a:n:c:tTid:Cxyz",["uuid=", "agency-id=", "agent-string=","cutoff=", "timestamp","friendly-timestamp","tripid","date=","currentday","print_driver","print_vehicle","print_id"])
```
unreasonableman, line 42: Thank you for fixing the inconsistent use of `-` and `_`!

unreasonableman (approved):

PR #77, opened 2022-02-01
Bugfix: filter short trips from GRaaS report

owades:
This is more cleanup of bugs created early in my GRaaS dev. I had filtered out trips shorter than 15min, but it turns out I had filtered based on schedule rather than GRaaS app sessions. This change removes the faulty logic, and builds new logic to filter out short GRaaS app trips - currently set to a 1min threshold but can be changed easily.

In order to do this, I update the Stats class to accept a list of doubles - and then generate that list before creating each new Stats class instance. This way, it can calculate min and max on two different values within GPSData.

A simpler approach could be to filter out trips with fewer than X location updates (regardless of duration). We could just count the number of items in the map and filter based on that. In my opinion, despite being more complicated, filtering based on duration is more intuitive.

In testing, this approach created identical GRaaS reports to what we are seeing on the main branch, with the very short trips filtered out. Example of two trips filtered out: note the trip with "000000000" is the result of having only one GPS point - and so that bug is addressed with this PR as well.
Screen Shot 2022-02-01 at 9 20 54 AM

### DIFF
--- a/gtfu/src/main/java/gtfu/GPSData.java
+++ b/gtfu/src/main/java/gtfu/GPSData.java
@@ -26,4 +26,12 @@ public class GPSData {
     public void increment() {
         count++;
     }
+
+    public Double getSecsSinceLastUpdateDouble() {
+        return Double.valueOf(secsSinceLastUpdate);
+    }
+
+    public Double getMillisDouble() {
+        return Double.valueOf(millis);
+    }
 }
--- a/gtfu/src/main/java/gtfu/GPSData.java
+++ b/gtfu/src/main/java/gtfu/GPSData.java
@@ -27,11 +27,11 @@ public class GPSData {
         count++;
     }
 
-    public Double getSecsSinceLastUpdateDouble() {
+    public Double getSecsSinceLastUpdateAsDouble() {
         return Double.valueOf(secsSinceLastUpdate);
     }
 
-    public Double getMillisDouble() {
+    public Double getMillisAsDouble() {
         return Double.valueOf(millis);
     }
 }
--- a/gtfu/src/main/java/gtfu/GPSData.java
+++ b/gtfu/src/main/java/gtfu/GPSData.java
@@ -1,6 +1,6 @@
 package gtfu;
 
-public class GPSData implements StatValue {
+public class GPSData {
     public long millis;
     public int secsSinceLastUpdate;
     public float lat;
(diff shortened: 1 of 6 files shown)


gtfu/src/main/java/gtfu/GPSData.java
```diff
@@ -27,11 +27,11 @@ public void increment() {
...
-    public Double getValue() {
-        if(secsSinceLastUpdate > 0) {
-            Double secsSinceLastUpdateDouble = Double.valueOf(secsSinceLastUpdate);
-            return secsSinceLastUpdateDouble;
-        } else return null;
+    public Double getSecsSinceLastUpdateDouble() {
+        return Double.valueOf(secsSinceLastUpdate);
+    }
+
+    public Double getMillisDouble() {
```
unreasonableman, line 34: Name should be `getMillisAsDouble `. Same for the method above.

gtfu/src/main/java/gtfu/GraphicReport.java
```diff
@@ -274,7 +275,11 @@ private void addRandomTestData(float prob) {
                     }
 
                     gpsMap.put(id,latLonMap);
-                    TripReportData td = new TripReportData(id, t.getHeadsign(), start, duration, "test", "test", "test", new Stats(gpsMap.get(id).values()));
+                    // Create lists as input to Stats
+                    List<Double> updateFreqList = gpsMap.get(id).values().stream().map(GPSData::getSecsSinceLastUpdateDouble).filter(secs -> secs > 0).collect(Collectors.toList());
```
unreasonableman, line 279: I'm not a fan of these constructs: they're long, convoluted and when you get an exception, it's difficult to see which part actually failed. But you built the whole design of this PR around them, so to not have them you would have to change all kinds of things.

unreasonableman (approved):

gtfu/src/main/java/gtfu/GraphicReport.java
owades: I wrote these originally as loops, and would be happy to revert. I'm curious for your take on which is preferable:

Previous code:
```            List<Double> updateFreqList = new ArrayList<>();
            List<Double> updateTimeList = new ArrayList<>();

            for (GPSData gpsData : gpsMap.get(id).values()) {

                if(gpsData.secsSinceLastUpdate > 0) {
                    Double secsSinceLastUpdateDouble = Double.valueOf(gpsData.secsSinceLastUpdate);
                    updateFreqList.add(secsSinceLastUpdateDouble);
                }

                updateTimeList.add(Double.valueOf(gpsData.millis));
            }```

unreasonableman:
I think the loop is preferable, but I admit that it's kind of a personal
preference/religious issue. Chaining is more modern java and has high
informational density, which can be viewed as a pro or con. As I said, if
you get an exception, the loop will let you understand more quickly what
has gone wrong.

PR #85, opened 2022-02-03
updates to triplist to help glendora & unitrans. updates to unitrans & glendora

owades:
You wouldn't think shape_id would give a viable name, but it was the best option for unitrans (with some editing necessary). Figured I'd keep it as an option for future agencies.


docs/server/onboarding-runbook.md
```diff
@@ -52,17 +52,17 @@ Create service data for agency
...
-- From `graas-staging/server/agency-config/gtfs`, run `./copy-to-bucket.sh <agency-id>` to copy the new data files to the project storage bucket (this script will throw an error and cancel upload if json is misformatted). Check the [storage bucket source of truth](https://console.cloud.google.com/storage/browser/graas-resources/gtfs-aux;tab=objects?project=lat-long-prototype&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false) to confirm that your updates went through. You can run this command whenever you update either file.
+- From `graas/server/agency-config/gtfs`, run `./copy-to-bucket.sh <agency-id>` to copy the new data files to the project storage bucket (this script will throw an error and cancel upload if json is misformatted). Check the [storage bucket source of truth](https://console.cloud.google.com/storage/browser/graas-resources/gtfs-aux;tab=objects?project=lat-long-prototype&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false) to confirm that your updates went through. You can run this command whenever you update either file.
```
unreasonableman, line 55: We don't want to put our actual google cloud project ID in here. Instead use something like '[YOUR_GOOGLE_CLOUD_PROJECTID]'

docs/server/onboarding-runbook.md
```diff
@@ -52,17 +52,17 @@ Create service data for agency
...
 - __id_ecdsa__: private key. This is __PRIVILEGED INFORMATION__ and needs to be kept confidential.
 - __id_ecdsa.pub__: public key
 
 - Move this new agency key folder to the graas-agency-keys repo (or elsewhere if you prefer). This is privleged information and shouldn't live on the open-source repo.
 
-- Create a unique agency QR code, which encodes the private key, to be read by the GRaaS app. Generate the code by going to the `graas-staging/server/qr-gen` directory and running `./run.sh ../../../graas-agency-keys/<agency-id>/id_ecdsa "<nickname for agency>"`. This will save a file called "qr.png" to the qr-gen folder - you'll want to paste this into the onboarding docs provided to the agency.
+- Create a unique agency QR code, which encodes the private key, to be read by the GRaaS app. Generate the code by going to the `graas/server/qr-gen` directory and running `./run.sh ../../../graas-agency-keys/<agency-id>/id_ecdsa "<nickname for agency>"`. This will save a file called "qr.png" to the qr-gen folder - you'll want to paste this into the onboarding docs provided to the agency.
```
unreasonableman, line 65: Please add some language explaining that the private key (and thus the QR code) is very sensitive information and should be treated accordingly. Optionally say something about how being careless with the QR code is equivalent to posting close-up pictures of your house keys on social media

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -336,13 +346,17 @@ public TripListEntry(TripData data, RouteData rdata, String firstStopID, String
...
 
+        // name = shapeID;
         if (nameField == "headsign"){
             name = headSign;
         } else if (nameField == "route_short_name") {
             name = routeShortName;
         } else if (nameField == "route_long_name") {
             name = routeLongName;
+        } else if (nameField == "shape_id") {
+            name = shapeID;
```
unreasonableman, line 359: You should *always* use `.equals()` for string comparisons instead of `==`. You can get lucky because of string internalization, but we should never count on that.

PR #89, opened 2022-02-08
resolve blank dropdown bug

owades:
The goal is to only show dropdown menus when they are ready (ie all trips, vehicles have loaded). Changes made:
- Show "loading GRaaS app" until all config files have loaded. At that point, show the dropdowns. This prevents the situation where a user taps "Load routes" before configuration has completed.
- loadTrips() now returns a map of trip info, in addition to setting a global var. populateTripList takes the map as a parameter. This way, we can run populateTripList(loadTrips()), to ensure synchronous behavior. This prevents a situation where populateTripList runs while trips are in the middle of re-loading.

- Set startMillis to Date.now() upon creation, since starting at 0 created an additional bug in which "Page will refresh" is repeatedly shown

Other non-bug-related changes:
- Replace the word "route" with "trip", to be more accurate to GTFS definitions
- Create hideElement(id) and showElement(id) functions, to simplify
- Some function descriptions
- Reduce dropdown width from 100% to 95%, so that they don't rub up against the right-hand side


unreasonableman (approved):
If you haven't please make to test on device also, and not just desktop browser.

server/app-engine/static/graas.js
```diff
@@ -731,9 +730,6 @@ function initialize() {
         version = version.substring(i + 2)
     }
     util.log("- version: " + version);
-    showElement(LOADING_TEXT_ELEMENT);
-    hideElement(START_STOP_BUTTON);
-    hideElement(ALL_DROPDOWNS);
```
owades: don't need to show loading here, and button + dropdowns are  hidden via HTML

PR #92, opened 2022-02-09
cleanup logging

owades:


unreasonableman (changes requested):
I should have been more specific. I meant: remove logging that creates a lot of output. Some logging is nice as a first step to understanding what is happening in case of issues.

The loadRoutes() stuff definitely needs to stay disabled, but please undo the other stuff and then put the app through its paces with the console open and see what either creates walls of text or triggers all the time. Anything else can stay.

+1 for bumping the version number!

PR #94, opened 2022-02-09
fix bug, remove logging

owades:
Prior to deploying last night I noticed a bug - the "Loading GRaaS" appears below the "Load trips" button for a split second after scanning QR code. This PR resolves that issue, and also does another pass at limiting logging

### DIFF
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -378,7 +378,6 @@ function handleStartStop() {
         util.log("+ delta: " + (millis - startMillis));
 
         hideElement(ALL_DROPDOWNS);
-
         showElement(LOADING_TEXT_ELEMENT);
 
         // Only load trips again if they were last loaded more than a minute ago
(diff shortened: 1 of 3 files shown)


server/app-engine/static/graas.js
```diff
@@ -752,18 +751,6 @@ function initialize() {
 function positionCallback() {
     util.log("+ got start position");
 
-    var dow = getDayOfWeek();
```
owades, line 755: I can't figure out the use for any of these

server/app-engine/static/gtfs-rt-util.js
```diff
@@ -134,21 +134,21 @@ if (!fetch) {
     exports.signAndPost = function(data, signatureKey, url, document) {
         this.log("util.signAndPost()");
         var data_str = JSON.stringify(data);
-        this.log('- data_str: ' + data_str);
-        this.log('- document: ' + document);
+        // this.log('- data_str: ' + data_str);
```
owades, line 137: the json still comes through in logs, this just removes a duplicate

unreasonableman (approved):

PR #95, opened 2022-02-10
GRaaS report updates in preparation for interactive GRaaS report

owades:
1. Accurate filenames for .png's (ie tcrta-2022-02-09)
2. Save all GRaaS reports to our storage bucket ([archive folder](https://console.cloud.google.com/storage/browser/graas-resources/graas-report-archive?project=lat-long-prototype&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)). Overwrite if it runs multiple times in a day.
3. If there are no reports on a given day, send an email that says "No reports today". (rather than 404 error)
4. Clean up downloadReport/savePath logic, to work as follows: 
      - If savePath is given, set downloadReport to true
      - If downloadReport is set to true, use a default savePath unless one is also given
      - Email sending is no longer tied to download report, so you have option to download report AND send email. To keep a consistent UX, and given that we generally want an email to send, a parameter only needs to be passed when you DON'T want an email sent

### DIFF
--- a/gtfu/src/main/java/gtfu/GraphicReport.java
+++ b/gtfu/src/main/java/gtfu/GraphicReport.java
@@ -80,11 +80,10 @@ public class GraphicReport {
     private int timeRowCount;
     private GCloudStorage gcs = new GCloudStorage();
 
-    public GraphicReport(String cacheDir, String selectedDate, boolean downloadReport, String savePath, boolean sendEmail) throws Exception {
+    public GraphicReport(String cacheDir, String selectedDate, String savePath, boolean sendEmail) throws Exception {
         Debug.log("GraphicReport.GraphicReport()");
         Debug.log("- cacheDir: " + cacheDir);
         Debug.log("- selectedDate: " + selectedDate);
-        Debug.log("- downloadReport: " + downloadReport);
         Debug.log("- savePath: " + savePath);
         Debug.log("- sendEmail: " + sendEmail);
 
@@ -145,7 +144,7 @@ public class GraphicReport {
             tdMap = dls.getTripReportDataMap();
             int startSecond = dls.getStartSecond();
 
-            if (downloadReport) {
+            if (savePath != null) {
                 String path = savePath + "/" + key;
                 Debug.log("- path: " + path);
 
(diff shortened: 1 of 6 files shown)


gtfu/src/main/java/gtfu/GraphicReport.java
```diff
@@ -195,17 +205,24 @@ public GraphicReport(String cacheDir, String selectedDate, String savePath, bool
...
-        SendGrid grid = new SendGrid(recipients, "Automated GRaaS Report", "Attached", blobs);
+        SendGrid grid = new SendGrid(recipients, "Automated GRaaS Report", "Attached", blobMap);
         int responseCode = grid.send();
     }
 
+    private void uploadToGCloud(String key, byte[] image) throws IOException {
+        // converts <agency-id>-yyyy-mm-dd to <agency-id>
+        String agencyID = key.substring(0, key.length() - 11);
+        String path = "graas-report-archive/" + agencyID;
+        gcs.uploadObject("graas-resources", path, key, image);
```
unreasonableman, line 223: if key is `<agency-id>-yyyy-mm-dd` where does the file name suffix, i.e. `png` come from?

gtfu/src/main/java/gtfu/GraphicReport.java
```diff
@@ -171,21 +176,26 @@ public GraphicReport(String cacheDir, String selectedDate, String savePath, bool
 
             // Only create report for agencies with trip report data
             if (tdList.size() > 0) {
-                blobs.add(imageToBlob(img));
+                // converts <agency-id>-yyyy-mm-dd.txt to <agency-id>-yyyy-mm-dd
+                blobMap.put(key.substring(0, key.length() - 4), imageToBlob(img));
             }
         }
-        if (downloadReport) {
```
unreasonableman, line 177: It doesn't seem like `downloadReport` really has a purpose anymore. Either `savePath` is `null`, which means don't download, or it has a value, which means download.

unreasonableman (approved):
Please see comment about arg redundancy

PR #98, opened 2022-02-10
resolve changes to sendgrid.java that served only graas report

owades:
Change broke LoadAgendyData test. Created an [issue](https://github.com/cal-itp/graas/issues/99) for SendGrid unit test.


unreasonableman (approved):
We are actually pretty close to making this class a very light-weight java API for sendgrid. It's kind of a separate project, but really what's missing is:
- changing name of map back to `blobMap`
- pass in a list of mime types instead of assuming `image\png`
- remove hardcoded sender, and instead use SendGrid.Builder approach to create and configure instances

Mostly just daydreaming, since we have all kinds of other stuff queued up

PR #106, opened 2022-02-18
interactive GRaaS report

owades:
Current version can be viewed [here](https://storage.googleapis.com/graas-resources/web/index.html).

Java updates include:
1. Updating GraphicReport to create a json file for each agency for every day, and save to gcloud.
2. Creation of various variables in GraphicReport
3. Removal of metadata text overlay
4. Creating AgencyListGenerator, which performs the specialized function of generating a JSON file to populate the agency and date select dropdowns (for each agency, logs which dates there is a graas report). AgencyListGenerator runs towards the end of GraphicReport.
5. Updates to gcloudstorage, mainly a new function which gets a list of objects in a given directory

Known issues:
- Scrolling is choppy in Safari, since it doesn't recognize the css "scroll-behavior" attribute

Upon merging this PR I'll need to:
1. Re-generate historic GRaaS reports, to generate JSON files and remove metadata (as an alternative we could also just delete historic reports and start fresh, I prefer the former).
2. Upload latest html, css & js to storage bucket (current version on bucket is based on some dev files

I considered updating the GRaaS Report email to contain a link, but held off since it's a chunk of work.


gtfu/src/web/graas-report-web.js
```diff
@@ -0,0 +1,343 @@
...
+                selectTrip(timelineCtx, trips[i].timeline);
+                scrollToTrip(trips[i])
+                break;
+            }
+        }
+    });
+}
+
+function objectContainsPoint(object, x, y){
+    if(x >= object.x){
```
unreasonableman, line 230: `return x >= object.x && x <= object.x + object.width && y >= object.y && y <= object.y + object.height;` is more succinct, functionally equivalent and won't cause branch prediction problems :)

unreasonableman: Also, it looks like you're not returning anything if any of the conditions fail. Maybe not returning anything in js is somehow the same as returning `false`, but this is poor style.

gtfu/src/web/graas-report-web.js
```diff
@@ -0,0 +1,343 @@
...
+
+    metadataWidth = trip.map.width;
+    metadataX = trip.map.x + metadataWidth;
+    metadataY = trip.map.y;
+    metadataHeight = trip.map.height;
+
+    while ((tooltipItems.length * (font_size + margin) >= metadataHeight)
+            || (getMaxTextWidth(trip) + margin * 2 >= metadataWidth))  {
+        console.log("reducing font size...");
+        font_size -= 5;
```
unreasonableman, line 260: What happens when the font size goes negative?

gtfu/src/web/graas-report-web.js
```diff
@@ -0,0 +1,343 @@
...
+    maxWidth = 0;
+    mapCtx.font= font_size + "px " + font;
+    for (var i = 0; i < tooltipItems.length; i++) {
+        text = tooltipItems[i] + ": " + trip[tooltipItems[i]];
+        length = mapCtx.measureText(text).width;
+        if (length > maxWidth){
+            maxWidth = length;
+        }
+    }
+    return Math.round(maxWidth);
```
unreasonableman, line 297: It's pedantery, but I think you want Math.ceil() here

unreasonableman (approved):
👍

gtfu/src/web/graas-report-web.js
owades: Since ceil() rounds down and round() rounds up, my understanding is that ceil() would result in the text width getting closer to the edge of the box. And round() would result in a bit more of a buffer.

gtfu/src/web/graas-report-web.js
owades: It shouldn't but that's a good question. I added some logic to simply not display the metadata if font size gets down to zero. This would occur if:
- Window is really small (possibly on mobile)
- Text is really really long

PR #111, opened 2022-03-01
tool to generate mappings from block IDs to trip IDs

unreasonableman:
Use like so: `java -cp build/lib/gtfu.jar gtfu.tools.BlockDataGenerator -c <cache-folder> -a <agency-id>`


owades (approved):
Looks good. I imagine in the future we might save it somewhere more specific than just printing.

PR #121, opened 2022-03-02
Automated updates to glendora-transportation-division triplist

owades:
Our daily check detected that changes were made to glendora-transportation-division's static GTFS. This PR was automatically generated, so please review and make updates if necessary before merging


owades:
The issue here is that I manually removed the word "Commuter Shuttle" when creating their file. It appears nothing else has changed, so we have the option to leave it as-is, use the longer name, or update triplistgenerator to exclude certain words or phrases.

PR #128, opened 2022-03-02
Automated trip-name checks & triplist improvements 

owades:
**UpdateTripNames**
- New file, UpdateTripNames, checks to see if static GTFS has been updated since the latest commit to trip-names.json. If so, regenerates the file and creates a PR ([example](https://github.com/cal-itp/graas/pull/136)) if it differs from the current one.
- Update "Load Agency Data Test" nightly GH action to also do the tripname check.
- Includes creation of GH.py, which uses a Java library to access the GH API to do various things on GH. Currently it uses my token, so the updates appear to come from me.

**TripListGenerator**
- Enable use of directionID, so agencies can use the route_name + direction_id combo (ie "Blue Line - Northbound"). Also add "dumb direction ID", which kicks in when agencies use direction_id in trips.txt but have no directions.txt file (Unitrans)
- Get triplistgenerator parameters (name and useDirectionID) from agency-params.json (renamed version of filter-params.json). 
- Create generateTripList() function so it can be run from another class
- Remove two unused components: "simple string" and "GTFSagency". I haven't been able to come up with a reason for keeping either (and Kay agreed to removing the former already).

**Storage bucket changes**
- Stop hosting agency-config json files on gcloud. Instead, point graas.js directly to github. This way a merged PR updates the source-of truth automatically.

**Misc**
- 2 updates to Util.py which break functions into sub-parts, so we can access the sub-part without creating redundancy
- Update GRaaS.js and index.html to refer to a trip as a "trip" rather than a "route".

Testing: I've been testing this via iPhone 12. Will need to do additional testing prior to shipping: trying to get the web app to run on my old Android now.


gtfu/src/main/java/gtfu/tools/GH.java
```diff
@@ -0,0 +1,89 @@
...
+import org.kohsuke.github.GHCommitQueryBuilder;
+import org.kohsuke.github.GHCommitBuilder;
+import org.kohsuke.github.GHTreeBuilder;
+import org.kohsuke.github.PagedIterable;
+import org.kohsuke.github.PagedIterator;
+import org.kohsuke.github.GHContentBuilder;
+
+import java.util.Date;
+
+public class GH {
```
unreasonableman, line 23: Terrible name! :P Please consider something GitHubUtil, GitHubShim or similar

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -268,24 +285,16 @@ public static void main(String[] arg) throws Exception {
...
-
-            if (gtfsAgency != null && !aid.equals(gtfsAgency)) {
-               list.remove(i--);
-            }
-        }
-
-        out.println();
+        // It's temping to replace the below manual JSON generation with JSONObject and JSONArray classes,
+        // but because they offer little control over formatting, this method is preferable for something
+        // needing to be read by humans.
```
unreasonableman, line 290: fasterxml's ObjectMapper has a method called writeValue() or something. While that doesn't support pretty-printing, you can get a pretty printer from the mapper and then call writeValue() on that. Not saying you have to, but you can if you want.

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -300,19 +309,30 @@ public static void main(String[] arg) throws Exception {
             out.close();
         }
     }
+
+    // For agencies who don't include the optional directions.txt file, assume the following mapping.
+    // This is designed specifically around Unitrans
```
unreasonableman, line 314: My recollection from reading the spec is that the meaning  of direction symbols is completely arbitrary. Please confirm. A better solution may be to make Unitrans add directions.txt

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -329,33 +349,28 @@ public RouteData(String agencyID, String routeShortName, String routeLongName) {
...
-            name = routeLongName;
+            name = rdata.routeLongName;
         } else if (nameField.equals("shape_id")) {
-            name = shapeID;
+            name = data.shapeID;
+        }
+        if (useDirection){
+            name += " - " + direction;
         }
         if (name == null || name == "") {
```
unreasonableman, line 375: As discussed, using the == operator on strings is a bad idea. Also, there is Util.isEmpty() which I think does both of the above checks. Please fix and also check for other occurrences of == with strings.

gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        if (prCount == 0) {
+            reporter.addLine("...no agencies. Everything looks up to date.");
+        }
+        reporter.send();
+    }
+
+    private static void usage() {
+        System.err.println("usage: UpdateTripNames [-u|--url <live-agencies-url>] [-a|--agency-id <agency-id>]");
+        System.err.println("    <url> is assumed to point to a plain text document that has an agency ID per line");
```
unreasonableman, line 90: <url> is not the name used two lines up. Please also add blurb for agency-id, and mention that either -u or -a must be given.

unreasonableman (changes requested):
Looks good mostly.

Make sure to address the string == comment, the rest is pretty much optional.

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
owades: Yes that's correct - definitely better for them to add directions.txt and I'll find a way to make that happen.

gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
owades: Updated the blurb but not sure what you mean by "not the name used 2 lines up"

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
owades: shame!

unreasonableman (approved):

PR #136, opened 2022-03-03
:robot: updates to glendora-transportation-division triplist

owades:
Our automated daily check detected that changes were made to glendora-transportation-division's static GTFS. This PR was automatically generated, so please review and make updates if necessary before merging


owades:
These changes are partially the result of taking away the max name length of 25 chars. We should test more, but I think it may be OK to do so.

owades:
After confirming that these longer names are still readable on my ios and android smartphones, here's how it looked on my pixel 2 from chrome:

Screenshot_20220303-140343

 I think it's OK to deploy this - with the value of reflecting the actual static GTFS outweighing the value of brevity.

unreasonableman (approved):

PR #137, opened 2022-03-03
use PST midnight timestamp in generating report name

owades:
Report was showing tomorrow's date in the GRaaS report title for agencies where the first trip started after midnight UTC (after 4pm pacific). See the attached file as an example.

Since we already do the work to calculate the timestamp of 12AM PST for that day, we should use that timestamp in producing the date - rather than using the first timestamp recorded by the agency that day.

This will be challenging to test without creating fake agency data (since our test agencies don't exist in agenies.yml and would thus fail). In testing thus far, I confirmed that it produces normal dates in the GRaaS report title when run for the current date and past dates. If we did want to test manually, the easiest way would be for me to post an update as Glendora at 4:30pm today, and if it showed up in the GRaaS report under today rather than tomorrow, we'd be good.

glendora-transportation-division-2022-02-24

### DIFF
--- a/.github/workflows/send-graas-report.yaml
+++ b/.github/workflows/send-graas-report.yaml
@@ -1,8 +1,6 @@
 name: Send GRaaS Report
 # ~9:18PM pacific time, daily:
-on:
- schedule:
-    - cron: '0 5 * * *'
+on: [push]
 jobs:
   Send-email:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


unreasonableman (approved):
I think your suggestion of testing by posting a manual update as glendora makes sense.

owades:
I ran the Glendora test - I wasn't able to reproduce the issue, until I realized that the issue only happens when GraphicReport is run via GH Actions. I tested the changes on github actions and confirmed that updates for all agencies are working as expected.

unreasonableman:
Digging deeper even when things seem to work. I like it!

PR #142, opened 2022-03-08
don't let log filter choke on json parsing issues

unreasonableman:
instead, show offending input


owades (approved):
Awesome! I know this issue

PR #144, opened 2022-03-08
Docs/initial java documentation

owades:
First round of using Javadoc, looking for a +1 on the general approach/level of detail.

Also removed startSecond from DayLogSlicer & GraphicReport as it's no-longer used.


gtfu/src/main/java/gtfu/tools/AgencyYML.java
```diff
@@ -29,6 +34,10 @@ public AgencyYML(){
         processYML();
     }
 
+    /**
+    * Identify the agencyName and GTFS schedule URL for every agency represented in the file.
+    * It is straightforward to add additional fields to pull from the file.
+    */
```
unreasonableman, line 40: Documenting private fields is really going above and beyond. We don't need to go beyond documenting public fields.

unreasonableman (approved):
Level of detail is definitely sufficient.

PR #145, opened 2022-03-09
pin itsdangerous to v2.0.1

owades:


owades:
fyi @unreasonableman - this did the track. (should have updated the previous PR verses merging and making this one)

PR #148, opened 2022-03-09
Cleanup following switch to github as agency json source of truth

owades:
Now that #128 is merged & deployed, some cleanup:
- Remove route-names and filter-params files, which have been replaced with trip-names and agency-params
- Remove copy-to-bucket.sh and copy-from-bucket.sh, as we no longer use the bucket for accessing these agency json files
- Update setup-agency-templates.sh to use new file names


owades:
@unreasonableman I updated the agency called "test", but not sure whether you still use it. Let me know if I should remove instead.

unreasonableman (approved):
Please keep test agency around.

PR #158, opened 2022-03-11
Set up bulk assignment mode (fka "patrol mode")

owades:
- Check whether agency's agency-config.json contains the optional use-bulk-assignment param (w/ a true value). If so, disable the trip selection dropdown.
- Report the trip-assignment-mode on the in-progress trip data view
- Begin caching vehicle ID, for 4 hours. If a driver stops a trip and starts again, regardless of bulk assignment mode, the vehicle-selection dropdown will auto-populate with the previously-selected value.
- A small amount of general cleanup

### DIFF
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -30,7 +30,7 @@ var startLon = null;
 var trips = [];
 var mode = 'vanilla';
 var sessionID = null;
-var useSimpleMode = false;
+var useBulkAssignmentMode = false;
 
 // Default filter parameters, used when agency doesn't have an agency-config.json file
 var maxMinsFromStart = 60;
@@ -400,7 +400,7 @@ function handleStartStop() {
             handleBusChoice();
         }
 
-        if(!useSimpleMode){
+        if(!useBulkAssignmentMode){
             // Only load trips again if they were last loaded more than a minute ago
             if ((millis - lastTripLoadMillis) < MILLIS_PER_MINUTE * 1) {
                 populateTripList();
@@ -497,7 +497,7 @@ function handleOkay() {
     document.cookie = `${VEHICLE_ID_COOKIE_NAME}=${vehicleID}; max-age=${MAX_VEHICLE_ID_AGE_SECS}`;
     util.log("- vehicleID: " + vehicleID);
 
-    if(!useSimpleMode){
+    if(!useBulkAssignmentMode){
         p = document.getElementById(TRIP_SELECT_DROPDOWN);
         var entry = tripIDLookup[p.value];
 
@@ -518,6 +518,10 @@ function handleOkay() {
 
     var p = document.getElementById('vehicle-id');
     p.innerHTML = "Vehicle ID: " + vehicleID;
+
+    var p = document.getElementById('trip-assignment-mode');
+    var tripAssignmentMode = (useBulkAssignmentMode ? 'bulk' : 'manual')
+    p.innerHTML = "Trip assignment mode: " + tripAssignmentMode;
 
     hideElement(ALL_DROPDOWNS);
     showElement(TRIP_STATS_ELEMENT);
@@ -859,7 +863,7 @@ function handleGPSUpdate(position) {
     data['vehicle-id'] = vehicleID;
     data['session-id'] = sessionID;
     data['pos-timestamp'] = posTimestamp;
-    data['use-simple-mode'] = useSimpleMode;
+    data['use-bulk-assignment-mode'] = useBulkAssignmentMode;
 
     if (driverName) {
         data['driver-name'] = driverName;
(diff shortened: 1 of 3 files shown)


unreasonableman:
> regardless of bulk assignment mode, the vehicle-selection dropdown will use the previously-selected value

When you say 'will use', you mean ' will pre-select', right? The driver still needs to confirm?

server/app-engine/static/graas.js
```diff
@@ -471,18 +493,22 @@ function handleOkay() {
...
+    if(!useBulkAssignmentMode){
+        p = document.getElementById(TRIP_SELECT_DROPDOWN);
+        var entry = tripIDLookup[p.value];
 
-    if (isObject(entry)) {
-        tripID = entry['trip_id'];
-    } else {
-        tripID = entry;
-    }
+        if (isObject(entry)) {
```
unreasonableman, line 504: looks like you have tabs or other extra indents here

server/app-engine/static/graas.js
```diff
@@ -966,14 +997,23 @@ function gotConfigData(data, agencyID, arg) {
             isFilterByDayOfWeek = data["is-filter-by-day-of-week"];
             maxMinsFromStart = data["max-mins-from-start"];
             maxFeetFromStop = data["max-feet-from-stop"];
+            if(data["use-bulk-assignment-mode"] != null){
```
unreasonableman, line 1000: !==

server/app-engine/static/graas.js
```diff
@@ -44,9 +45,12 @@ var testDow = null;
 var version = null;
 
 const UUID_NAME = 'lat_long_id';
+const VEHICLE_ID_COOKIE_NAME = 'vehicle_id';
+var vehicleIDCookie = null;
 const MSG_NO = 'msg_no';
-const MAX_AGE = 10 * 60 * 60 * 24 * 365;
-const MAX_LIFE = 1000 * 60 * 60 * 24 * 7;
+const MAX_AGE_SECS = 10 * 60 * 60 * 24 * 365;
```
unreasonableman, line 51: You're playing a somewhat dangerous game here as long as we don't have a linter. It'd be pretty easy to overlook an instance or two where the name needs to be changed. I understand your desire for clean code, and I'm not saying don't do it, just be aware.

unreasonableman (approved):

owades:
> > regardless of bulk assignment mode, the vehicle-selection dropdown will use the previously-selected value
> 
> When you say 'will use', you mean ' will pre-select', right? The driver still needs to confirm?

@unreasonableman correct - it will be selected and the driver still needs to tap "start trip". I'll update the comment

server/app-engine/static/graas.js
owades: Thanks. I just double checked for other files using these variables. It's a helpful reminder, can be easy to forget that these vars could be referenced in other files.

server/app-engine/static/graas.js
owades: don't think so, might be missing something

owades:
In testing this PR I've started seeing a new bug, very intermittently: "Unhandled promise rejection...". I need to figure out how to reproduce and resolve this before merging.

Mostly recently the error occurred when I tried to load the GRaaS app as Glendora, who does not have the "bulk assignment" variable in their config json file. Upon rebooting the local server, it loaded fine.

Screen Shot 2022-03-11 at 10 14 39 AM

owades:
I haven't been able to reproduce the bug (except very intermittently on Monday and Tuesday), after trying across multiple browsers and devices. I also haven't been able to find similar accounts of intermittent success in importing a key. 

I can reproduce the error by changing `i < strLen` to `i < strLen - 1`in the str2ab function:
```
    exports.str2ab = function(str) {
        const buf = new ArrayBuffer(str.length);
        const bufView = new Uint8Array(buf);

        for (let i = 0, strLen = str.length; i < strLen; i++) {
            bufView[i] = str.charCodeAt(i);
        }

        return buf;
    };
```

I pivoted to handling the error rather than preventing it. When the error occurs, the following happens:
- Delete the key from local storage
- Present an error message
- Automatically refresh the page

The only downside to this approach is that the QR Reader tool requests camera permissions again - I don't see this as a huge deal worth addressing. Screenshot of error, from iPad:
IMG_A210169269F7-1

server/app-engine/static/graas.js
```diff
@@ -471,18 +491,22 @@ function handleOkay() {
     p.style.background = "green";
     var p = document.getElementById(BUS_SELECT_DROPDOWN);
     vehicleID = p.value
+
+    document.cookie = `${VEHICLE_ID_COOKIE_NAME}=${vehicleID}; max-age=${MAX_VEHICLE_ID_AGE_SECS}`;
```
unreasonableman, line 495: Did you test that the cookie expires at max-age?

owades: Yep - tested on a MAX_VEHICLE_ID_AGE_SECS of 5 and confirmed it works as expected

unreasonableman (approved):

PR #160, opened 2022-03-11
:robot: updates to unitrans triplist

owades:
Our automated daily check detected that changes were made to unitrans's static GTFS. This PR was automatically generated, so please review and make updates if necessary before merging


owades:
In manual review (file too big for diff), it looks like the updates are just a changing of the order of trips. Going to merge.

owades:
In manual review (file too big for diff), it looks like the updates are just a changing of the order of trips. Going to merge.

PR #164, opened 2022-03-16
Update agency-params.json for tcrta

owades:


server/agency-config/gtfs/gtfs-aux/tcrta/agency-params.json
```diff
@@ -1,7 +1,7 @@
 {
 "is-filter-by-day-of-week": true,
 "max-mins-from-start": 45,
```
unreasonableman, line 3: As I just said in my email reply, I would suggest turning off all filtering for now to confirm it resolves the issue.

PR #167, opened 2022-03-17
bulk assignment updates for server and dispatch UI

unreasonableman:
- "working" indicator
- more dynamic layout to accommodate different screen sizes and aspect ratios
- "dirty" state
- guard against metadata blocks with older format
- add scaffolding for sound playback (not yet working)
- add toast mechanism to display UI
- use toasts to confirm which blocks were unassigned with either long press or by pressing close button
- fix bug that passed wrong x, y coordinates to longPress() function
- fix bug where timers weren't being canceled
- update "dirty" state when unassiging blocks


owades (approved):
Looks good!

PR #180, opened 2022-03-18
bug fix and minor UI cleanup

unreasonableman:
- fix a couple of "dirty" state bugs
- rounded corners and box shadow for toast and modals

### DIFF
--- a/server/app-engine/static/vehicle-assignments.js
+++ b/server/app-engine/static/vehicle-assignments.js
@@ -601,7 +601,7 @@ function handleCloseButtonPress() {
         }
 
         for (var i=currentAssignments.length-1; i>=0; i--) {
-            if (currentAssignments[i].blockID === closeButtonData.item.label) {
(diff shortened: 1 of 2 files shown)


server/app-engine/static/vehicle-assignments.js
```diff
@@ -624,7 +624,7 @@ function handleCloseButtonPress() {
         }
 
         for (var i=currentAssignments.length-1; i>=0; i--) {
-            if (currentAssignments[i].vehicleID === closeButtonData.item.vehicle) {
+            if (currentAssignments[i].blockID === closeButtonData.item.label) {
                 currentAssignments.splice(i, 1);
                 break;
             }
```
owades, line 630: Looking at this and the above change... blockID and vehicleID can both be equal to a label value? Doesn't seem like an issue (since both values appear on a button) but just double checking

unreasonableman: Yes, they'r all strings

PR #182, opened 2022-03-18
Add filtering for calendars.txt start_date and end_date to web app

owades:
- Filter based on logic in GTFS spec
- Filters can be overrided by ignore-start-end-date in agency-params.json
- Test value can be passed in via utm param: testdate
- This PR also removes two unused features: driver-select dropdown and "mode"

For a test agency, I confirmed:
- Filtering will show a trip if the current date is the start date or end date
- Filtering will not show a trip if current date is before start, or after end
- Filtering will show all trips if ignore-start-end-date is set to true (though, cache doesn't refresh immediately)

### DIFF
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -992,7 +992,7 @@ function loadTrips() {
                 //util.log(`- time: ${time}`);
                 const dow = util.getDayOfWeek();
                 //util.log(`- dow: ${dow}`);
-                const date = util.getDate();
+                const date = util.getTodayYYYYMMDD();
                 util.log(`- date: ${date}`);
                 const lat = tripInfo.departure_pos.lat;
                 // util.log(`- lat: ${lat}`);
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -23,8 +23,6 @@ var lastTripLoadMillis = 0;
 var configMatrix = null;
 var countFiles = null;
 var countGetURLCallbacks = 0;
-var driverList = null;
-var driverName = null;
 var startLat = null;
 var startLon = null;
 var trips = [];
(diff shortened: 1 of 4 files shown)


server/app-engine/static/graas.js
```diff
@@ -90,6 +87,19 @@ function getDayOfWeek() {
...
+        return testDate;
+    } else {
+        var today = new Date();
+        var year = today.getFullYear();
+        var month = today.getMonth() + 1;
+        var date = today.getDate();
+        return (year * 10000) + (month * 100) + date;
+    }
+}
+
```
unreasonableman, line 102: gtfs-rt-util.js has some time and date utility functions. Consider adding this there.

unreasonableman (approved):

server/app-engine/static/graas.js
owades: updated - mind taking one more quick look? Tested and is working fine

server/app-engine/static/gtfs-rt-util.js
```diff
@@ -89,6 +89,63 @@ if (!fetch) {
...
+
+    // returns 0 for Monday...and 6 for Sunday
+    exports.getDayOfWeek = function() {
+        if (testDow) {
+            return testDow;
+        } else return ((new Date()).getDay() + 6) % 7;
+    }
+
+    // returns date as an 8-character string (ie 20220317 for 3/17/22)
+    exports.getDate = function() {
```
unreasonableman, line 101: gtfs-rt-util.js already has a getDate() function (that does something different). Choose a different name, or else one function will be permanently hidden.

unreasonableman (changes requested):
please see comment

unreasonableman (approved):

PR #185, opened 2022-03-21
bulk assignment bug fixes

unreasonableman:
- fix crash bug in BlockDataGenerator for absolute dates
- make server dates timezone aware
- fix timezone bug
- fix erratic zoom behavior by disabling pinch-to-zoom


owades (approved):
Looks good. I checked where get_current_time_millis() is used, and can confirm that it looks like the changes to the function don't appear to have downstream consequences.

PR #189, opened 2022-03-23
js linting

unreasonableman:
- create a common node folder: server/node
- move node-fetch package to that folder
- install needed packages for eslint
- create eslint standard config file and start customizing it
- create script to run eslint over all client-side js files: server/tools/eslint/lint-files.sh (currently generates >10K errors)
- we need to discuss the set of linting rules we will adopt and enforce


owades (approved):
Looking forward to reviewing errors and setting up rules.

evansiroky:
oof, you definitely don't want to be checking in a `node_modules` folder into a repo. See [this article](https://sebhastian.com/git-ignore-node_modules/) for more info.

PR #202, opened 2022-03-29
add pr test to check for pos update in db

owades:


gtfu/src/test/java/gtfu/test/ServerTest.java
```diff
@@ -120,22 +121,46 @@ private static boolean checkForServiceAlertFeed(String agency) throws Exception
...
+        int searchWindowSecs = 30;
+
+        DB db = new DB();
+        long nowSecs = Util.now() / 1000;
+        long queryStartTime = nowSecs - searchWindowSecs;
+        long queryEndTime = nowSecs;
+        List<String> results = db.fetch(queryStartTime, queryEndTime, "position", propertyNames);
+
+        for (int i=0; i < results.size(); i++){
+            if(results.get(i).equals(agency)){
```
unreasonableman, line 137: The assumption here is that no one else is submitting updates as agency `test`, which is of course not guaranteed. Agency name `test` is currently hardcoded in post-position-updates.js. It would be nice if a) there was a way to pass an arg when invoking this (whether it's agency id or something else) and then b) pass a randomly generated value and check for exactly that value in the code above.

unreasonableman: I misread the test code. Agency id is set to `pr-test`. not `test`. Could still come from somewhere else, but much less likely. So, feel free to disregard.

unreasonableman (approved):

gtfu/src/test/java/gtfu/test/ServerTest.java
owades: I think the approach you described would be more watertight, and that existing will work well for now

PR #205, opened 2022-03-30
Bugfix/update dependencies

owades:
web app deploy had the error:
```
ImportError: cannot import name 'escape' from 'jinja2' (/layers/google.python.pip/pip/lib/python3.7/site-packages/jinja2/__init__.py)
```

Updating flask did the trick - and the front end (viewable [here](https://20220329t215335-dot-lat-long-prototype.wl.r.appspot.com/)) is looking good.

The next step will be to migrate traffic later tonight and then perform backend tests (or to continue looking into how to perform backend tests on a single version)

### DIFF
--- a/server/app-engine/requirements.txt
+++ b/server/app-engine/requirements.txt
@@ -9,3 +9,4 @@ ecdsa==0.16.0
 google-cloud-ndb==1.11.1
 google-cloud-storage==1.43.0
 itsdangerous==2.0.1
+pytz==2021.1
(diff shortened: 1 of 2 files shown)


unreasonableman (approved):
As just mentioned in the chat, another option could be to move all contents of `templates` to `static`, since placing stuff in the former is telling flask to run jinja on it. We're not currently doing any templating.

owades:
Let's try that approach as it simplifies things a bit. We will need to upgrade flask sooner or later, though.

PR #206, opened 2022-03-30
:robot: updates to unitrans triplist

owades:
Our automated daily check detected that changes were made to unitrans's static GTFS. This PR was automatically generated, so please review and make updates if necessary before merging


owades:
@unreasonableman the diff actually loads for this one, and everything looks normal. Usually I merge these auto-generated PRs without review but I want to give you a chance to review Unitrans.

unreasonableman (approved):
Looks reasonable upon cursory inspection.

There are often 5 - 10 trips with the same name and departure time, but they are distinct in either DOW or cal start and end date from what I can see.

Doing this inspection manually is tedious and error prone, as we discussed yesterday. We want to automatically check that:
- the JSON parses
- each entry has the number of fields we expect
- that the number of entries stays roughly the same (say, within 25% of the previous number of entries). If a change lies outside that boundary, generate a note/warning

PR #223, opened 2022-04-01
Refactor TripListGenerator to use Util.LoadCollections

owades:
Refactor TLG to use existing functions - and add bits of missing functionality to Util.LoadCollections as needed. Also add Direction and DirectionCollection classes.

I ran UpdateTripNames on all live agencies for a final diff comparison, and the only difference is some rounding on lat and long points. Looks like the result of Float.parseFloat() in StopCollection.java, but I'm inclined to ignore this because: 1) The difference is a matter of feed and 2) It should be now using lat and long values that match the rest of GTFU standard classes. You can see these diffs as they are all sitting in PR drafts.

Separately, I also updated Block Generator to use Util.updateCacheIfNeeded(), as it appears it was using an expired cache. I tested this by re-running the nightly block generation job.

### DIFF
--- a/gtfu/src/main/java/gtfu/Calendar.java
+++ b/gtfu/src/main/java/gtfu/Calendar.java
@@ -1,6 +1,7 @@
 package gtfu;
 
 import java.util.Date;
+import java.util.Arrays;
 
 public class Calendar {
     public static final int MONDAY    = java.util.Calendar.MONDAY;
@@ -15,11 +16,15 @@ public class Calendar {
 
     private String serviceID;
     private int days[];
+    private String startDate;
+    private String endDate;
     private long startMillis;
     private long endMillis;
 
-    public Calendar(String serviceID, int monday, int tuesday, int wednesday, int thursday, int friday, int saturday, int sunday, long startMillis, long endMillis) {
+    public Calendar(String serviceID, int monday, int tuesday, int wednesday, int thursday, int friday, int saturday, int sunday, String startDate, String endDate, long startMillis, long endMillis) {
         this.serviceID = serviceID;
+        this.startDate = startDate;
+        this.endDate = endDate;
         this.startMillis = startMillis;
         this.endMillis = endMillis;
 
@@ -51,6 +56,14 @@ public class Calendar {
         return days[dow] != 0;
     }
 
+    public String getStartDate() {
+        return startDate;
+    }
+
+    public String getEndDate() {
+        return endDate;
+    }
+
     public String toShortList() {
         StringBuffer sb = new StringBuffer();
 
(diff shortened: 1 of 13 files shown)


gtfu/src/main/java/gtfu/Calendar.java
```diff
@@ -71,4 +83,27 @@ public String toShortList() {
...
+        4 -> 6
+        5 -> 7
+        6 -> 1
+        */
+
+        sb.append("[");
+        for (int i=0; i<7; i++) {
+            int n = ((i + 1) % 7 + 1);
+            sb.append(days[n] == 0 ? "0" : "1");
+            sb.append(i < 6 ? ", " : "");
```
unreasonableman, line 104: May I suggest `DAYS_PER_WEEK - 1` here and `DAYS_PER_WEEK` in the two occurrences above? Should probably go with the other constants in `Time.java`.

gtfu/src/main/java/gtfu/Direction.java
```diff
@@ -0,0 +1,65 @@
...
+    public Direction(String routeDirectionID, String routeID, String directionID, String name) {
+        this();
+
+        this.routeDirectionID = routeDirectionID;
+        this.routeID = routeID;
+        this.directionID = directionID;
+        this.name = name;
+    }
+
+    private Direction() {
```
unreasonableman, line 28: the private constructor and the serialization methods `read()` and `write()` are relics of a time when I thought we would have to constantly run the prediction engine on always-on servers. This thankfully turned out to be false, and thus we shouldn't have these constructs anymore.

gtfu/src/main/java/gtfu/DirectionCollection.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private DirectionCollection() {
+        map = new HashMap<String, Direction>();
+        list = new ArrayList<Direction>();
+    }
+
+    public void write(DataOutputStream out) {
+        try {
+            out.writeInt(list.size());
+            int count = 0;
```
unreasonableman, line 66: unused variable

unreasonableman (approved):
👍 for addressing this cruft!

If you don't feel like addressing the serialization issue mentioned in the comments right now, please add a GH issue for it.

gtfu/src/main/java/gtfu/DirectionCollection.java
```diff
@@ -0,0 +1,77 @@
...
+                Direction direction = new Direction(routeDirectionID, routeID, directionID, name);
+
+                map.put(routeDirectionID, direction);
+                list.add(direction);
+            }
+
+            tf.dispose();
+        }
+        catch (Exception e){
+            e.printStackTrace();
```
owades, line 49: @unreasonableman I changed this from Util.Fail() to just a stacktrace, as this fails when an optional file is missing (ie not a big ideal). It was printing errors on the daily emails (ie block generation) so I think it's not a good use use of Fail (and this was with skipErrors hardcoded to true)

PR #224, opened 2022-04-01
don't persist node-modules folder

unreasonableman:
Requested by @evansiroky, some details [here](https://sebhastian.com/git-ignore-node_modules/).

Re-generate node-modules with `npm install` issued in the folder that has packages.json.


unreasonableman:
@owades hold on, I have a failed check.

unreasonableman:
@owades fixed `pr-test` failure, pls review

.github/workflows/pr-test.yaml
```diff
@@ -27,6 +27,9 @@ jobs:
           # but this method is continuing to be supported. More info: https://github.com/google-github-actions/auth#setup
           service_account: 'lat-long-prototype@appspot.gserviceaccount.com'
           credentials_json: ${{ secrets.GCP_SA_KEY_JSON }}
+      - name: Install node packages
+        working-directory: server/node/
+        run: npm install
```
evansiroky, line 32: Another option here is to use the [setup-node](https://github.com/actions/setup-node) GitHub action. It has built-in caching of the installed packages.

owades (approved):

.github/workflows/pr-test.yaml
unreasonableman: Isn't that what line 20 does?

evansiroky: Ah, didn't see that in there. Still, according to the [docs](https://github.com/actions/setup-node#caching-packages-dependencies), caching is turned off by default, so enabling it would probably speed up these builds slightly.

PR #226, opened 2022-04-05
Prevent TLG crash on github actions when creating new file

owades:
TripListGenerator was crashing on github actions because it wasn't able to handle the filepath: System.getenv("HOME") + "/tmp/tuff". I updated the default path, and tested the fix both locally and on GH actions.

Since I performed all PR testing last friday locally rather than on Github Actions, I didn't catch this issue before merging.

Originally I was thinking I would add some logic that prevents UpdateTripNames from crashing when TLG fails for a single agency, but I had a bit of trouble coming up with a solution and not sure it is necessary right now.

Lots of commits since I was using the "on push" github action trigger for testing.

### DIFF
--- a/.github/workflows/daily-static-gtfs-check.yaml
+++ b/.github/workflows/daily-static-gtfs-check.yaml
@@ -30,7 +30,7 @@ jobs:
           SENDGRID_API_KEY: ${{secrets.SENDGRID_API_KEY}}
       - name: Update Trip Names
         working-directory: gtfu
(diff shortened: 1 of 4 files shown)


unreasonableman (approved):
This doesn't solve the general problem of how to keep one error from torpedoing the entire action, which is sure to occur again.

owades:
Created https://github.com/cal-itp/graas/issues/229 as followup

PR #227, opened 2022-04-05
Add dates to block-generation email

owades:
Email now shows which dates have new block data generated for each agency:
Screen Shot 2022-04-05 at 3 44 19 PM

I tested by running BulkBlockGenerator via GH actions (email pasted above) as well as BlockGenerator from command line.

### DIFF
--- a/.github/workflows/daily-static-gtfs-check.yaml
+++ b/.github/workflows/daily-static-gtfs-check.yaml
@@ -1,8 +1,8 @@
 name: Daily Static GTFS Processes
-on: [push]
+on:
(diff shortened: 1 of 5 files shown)


unreasonableman:
It looks like we are generating three day's worth of block data for an agency at a time through the action. Is the action only called every three days, or are we regenerating data that already exists, or am I missing something?

unreasonableman (approved):

owades:
That's correct - currently it generates 3 days out, and overwrites when it already exists. This can be modified.

PR #228, opened 2022-04-05
Create python docstrings for server code, integrate into mkdocs

unreasonableman:
Requires the following pip installs:
- mkdocstrings
- mkdocs
- mkdocs-material

I haven't added these to requirements.txt, as otherwise they would be installed in app-engine for deploys. Open to ideas on how to better handle these doc dependencies.

### DIFF
--- a/docs/python-docs.md
+++ b/docs/python-docs.md
@@ -1,0 +1,2 @@
+::: server.app-engine
+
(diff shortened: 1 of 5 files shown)


owades:
@unreasonableman looking through this now - is there somewhere I can see the mkdocs output, or does it need to be deployed first?

unreasonableman:
You can do `mkdocs serve` from the root folder, which starts up a local web server.

owades (approved):
looks good, assuming you've tested it out on local server

PR #233, opened 2022-04-08
Auto-generate block data for test agency

owades:
I thought through a couple options here, and went with the one that didn't involve adding "test" to live-agencies.txt, and involved minimal hardcoding of test configuration. My thought process is that we want that file to be a source of truth for actual live agencies, and adding test values could cause unexpected consequences.


unreasonableman (changes requested):
Please use https://github.com/cal-itp/graas/blob/features/block-assignment-ui-polish/gtfu/src/main/java/gtfu/tools/BlockDataGenerator.java. That way, nothing needs to be hardcoded, which means it's more versatile.

Then just change  line 48 in daily-static-gtfs-check.yaml to include the `--url \<unitrans-gtfs-url\>` arg and we should be good to go.

owades:
@unreasonableman ready for re-review. I'm not sure if I'm just a bit slow after the weekend, but I didn't realize gtfsURL was already a parameter in main()  on the main branch. So the diff has very limited updates!

unreasonableman (approved):

PR #234, opened 2022-04-08
bulk assignment UI work

unreasonableman:
- clip labels that are too long and add ellipses, show full label on hover/long press
- show a friendly description for blocks upon click/long press (block ID and then a list of head signs, each with a list of departure times) 
- works for both desktop and mobile

A couple of items that snuck in but aren't actually bulk assignment UI:
- don't choke on missing `directions.txt` in `Util.loadCollections()`
- allow `BlockDataGenerator` to work on agencies not present in agencies.yaml

### DIFF
--- a/gtfu/src/main/java/gtfu/tools/BlockDataGenerator.java
+++ b/gtfu/src/main/java/gtfu/tools/BlockDataGenerator.java
@@ -78,7 +78,7 @@ public class BlockDataGenerator {
                 Trip trip = trips.get(tripID);
 
                 tr.id = tripID;
-                tr.headSign = trip.getHeadsign();
+                tr.headsign = trip.getHeadsign();
                 tr.startTime = trip.getStartTime();
                 tr.endTime = tr.startTime + trip.getDurationInSeconds();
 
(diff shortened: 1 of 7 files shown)


gtfu/src/main/java/gtfu/tools/BlockDataGenerator.java
```diff
@@ -121,8 +124,9 @@ public BlockDataGenerator(String cacheFolder, String outputFolder, String agency
     }
 
     private static void usage() {
-        System.err.println("usage: BlockDataGenerator -a|--agency-id <agency-id> -u|--upload [-c|--cache-folder <cache-folder>] [-o|--output-folder <output-folder>] [-d|--date <mm/dd/yy>|<n>]");
+        System.err.println("usage: BlockDataGenerator -a|--agency-id <agency-id> [-U|--url <static-gtfs-url>] -u|--upload [-c|--cache-folder <cache-folder>] [-o|--output-folder <output-folder>] [-d|--date <mm/dd/yy>|<n>]");
         System.err.println("    <agency-id> a transit agency identifier constructed from the alphabet of [a-z\\-]");
+        System.err.println("    <static-gtfs-url> link to agency's static static GTFS data. System attempts to backfill with data from agency.yaml if omitted");
```
owades, line 129: How about:
Optional link to agency's static static GTFS data. For agencies present in agencies.yml, the URL will be loaded automatically

gtfu/src/main/java/gtfu/tools/BlockDataGenerator.java
```diff
@@ -188,11 +198,12 @@ public static void main(String[] arg) throws Exception {
 
         if (agencyID == null) usage();
 
-        new BlockDataGenerator(cacheFolder, outputFolder, agencyID, date, uploadToGcloud);
+        new BlockDataGenerator(cacheFolder, outputFolder, agencyID, gtfsURL, date, uploadToGcloud);
     }
 
     class TripRecord {
         public String id;
+        public String headSign;
```
owades, line 206: very minor but it's just one one word, so camelcase is unnecessary. Just "headsign"

owades (approved):

server/app-engine/static/gtfs-rt-util.js
```diff
@@ -111,7 +122,7 @@ if (!fetch) {
     }
 
     // 's' is assumed to be a time string like '8:23 am'
-    function getTimeFromString(s) {
+    exports.getTimeFromString = function(s) {
```
owades, line 125: @unreasonableman do you know what the reason for this change was? I should have caught this before - I had written it this way deliberately since the function is only used within gtfs-rt-util.js and doesnt need be exported.

unreasonableman:
I just saw it stick out. Didn't realize it was only used internally.

PR #238, opened 2022-04-13
Prevent TripListGenerator from breaking nightly github action

owades:
...and fix NullPointerException issue that occurred last night.

I'm marking this as resolving #229, since it catches and reports issues in TLG, and TLG is what has caused the action to fail the most. I tried a more robust way of having each java tool run independently, but I believe doing so would require having them run independently of important upstream steps as well.

### DIFF
--- a/gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
+++ b/gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
@@ -66,6 +66,9 @@ public class UpdateTripNames {
                 String utf8 = StandardCharsets.UTF_8.name();
                 try (PrintStream ps = new PrintStream(baos, true, utf8)) {
                     TripListGenerator.generateTripList(agencyID, null, ps, false);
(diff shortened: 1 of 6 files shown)


gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
```diff
@@ -66,6 +66,9 @@ public static void UpdateTripNames(String[] agencyIDList) throws Exception {
                 String utf8 = StandardCharsets.UTF_8.name();
                 try (PrintStream ps = new PrintStream(baos, true, utf8)) {
                     TripListGenerator.generateTripList(agencyID, null, ps, false);
+                } catch (Exception e) {
+                    reporter.addLine("   * exception in generating triplist for " + agencyID);
```
unreasonableman, line 70: Might as well add the exception to the error line, e.g. `  * exception generating triplist for " + agencyID + ": " + e`

unreasonableman (approved):

PR #241, opened 2022-04-14
fix some JS issues, including stuck dispatch UI on mobile

unreasonableman:
- fix long press race condition in dispatch UI, as reported by @owades
- export util time constants
- fix internal call of getTimeFromString()

tested grass client and dispatch UI on mobile


unreasonableman:
pr-test failed, please stand by

owades (approved):

PR #242, opened 2022-04-15
setup eslint and perform initial set of fixes

owades:
- Update eslint to not fail for currently common issues, and to run as a part of PR test. 
- Make updates mainly to graas.js, gtfs-rt-util.js and vehicle-assignments.js, in order to pass the no-redeclare test. One of the bigger changes here is introducing let instead of var
- Note that I haven't yet figured out how to have eslint catch the duplicate variable issue that was resolved in #241 - still working at it.

I would appreciate a detailed look at the vehicle-assignments.js changes, as this is not code I'm completely familiar with.

### DIFF
--- a/server/app-engine/static/graas.js
+++ b/server/app-engine/static/graas.js
@@ -85,15 +85,15 @@ function toRadians(degrees) {
 
 // return distance in feet between to lat/long pairs
 function getHaversineDistance(lat1, lon1, lat2, lon2) {
-    let phi1 = toRadians(lat1);
-    let phi2 = toRadians(lat2);
-    let deltaPhi = toRadians(lat2 - lat1);
-    let deltaLam = toRadians(lon2 - lon1);
+    const phi1 = toRadians(lat1);
+    const phi2 = toRadians(lat2);
+    const deltaPhi = toRadians(lat2 - lat1);
+    const deltaLam = toRadians(lon2 - lon1);
 
-    let a = Math.sin(deltaPhi / 2) * Math.sin(deltaPhi / 2)
+    const a = Math.sin(deltaPhi / 2) * Math.sin(deltaPhi / 2)
         + Math.cos(phi1) * Math.cos(phi2)
         * Math.sin(deltaLam / 2) * Math.sin(deltaLam / 2);
-    let c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
+    const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
     return EARTH_RADIUS_IN_FEET * c;
 }
 
@@ -473,7 +473,7 @@ function getCookie(name) {
     const cookie = document.cookie;
     if (!cookie) return null;
 
-    let list = cookie.split('; ');
+    const list = cookie.split('; ');
 
     for (let i=0; i<list.length; i++) {
         if (list[i].startsWith(name + '=')) {
@@ -590,15 +590,15 @@ function handleEvent(e) {
 }
 
 function getRewriteArgs() {
-    let s = window.location.search;
-    let arg = s.split(/[&\?]/);
+    const s = window.location.search;
+    const arg = s.split(/[&\?]/);
 
     for (let a of arg) {
         if (!a) continue;
 
-        let t = a.split('=');
-        let key = t[0];
-        let value = t[1];
+        const t = a.split('=');
+        const key = t[0];
+        const value = t[1];
 
         util.log(`- ${key}: ${value}`);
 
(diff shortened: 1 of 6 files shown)


server/app-engine/static/vehicle-assignments.js
```diff
@@ -340,22 +344,22 @@ function handleKey(id) {
         util.log('- fromDate: ' + fromDate);
         util.log('- toDate  : ' + toDate);
 
-        for (var item of items) {
+        for (item of items) {
             if (item.type === 'block' && item.status === 'assigned') {
-                var id = item.label;
-                util.log('-- id: ' + id);
+                var itemID = item.label;
```
owades, line 349: Please give a good look at this line - I'm renaming "id" to "itemID". Note that id is also a param for this function, and this change assumes we don't want that parameter value stored in the itemID variable. Is that correct?

server/app-engine/static/vehicle-assignments.js
```diff
@@ -340,22 +342,22 @@ function handleKey(id) {
         util.log('- fromDate: ' + fromDate);
         util.log('- toDate  : ' + toDate);
 
-        for (var item of items) {
+        for (item of items) {
```
unreasonableman, line 345: Why is there no `let` here?

server/app-engine/static/vehicle-assignments.js
```diff
@@ -340,22 +342,22 @@ function handleKey(id) {
...
             if (item.type === 'block' && item.status === 'assigned') {
-                var id = item.label;
-                util.log('-- id: ' + id);
+                var itemID = item.label;
+                util.log('-- itemID: ' + itemID);
                 var vid = item.vehicle;
                 util.log('-- vid: ' + vid);
-                var block = blockMap[id];
+                var block = blockMap[itemID];
```
unreasonableman, line 351: Shouldn't we use `let` everywhere instead of `var` for local variables?

server/app-engine/static/vehicle-assignments.js
```diff
@@ -641,7 +643,7 @@ function handleCloseButtonPress() {
     var blockID = null;
 
     if (closeButtonData.item.type === 'vehicle') {
-        for (var item of items) {
+        for (item of items) {
```
unreasonableman, line 646: Same question as above

server/app-engine/static/vehicle-assignments.js
```diff
@@ -653,7 +655,7 @@ function handleCloseButtonPress() {
             }
         }
 
-        for (var i=currentAssignments.length-1; i>=0; i--) {
+        for (i=currentAssignments.length-1; i>=0; i--) {
```
unreasonableman, line 658: Ditto. Please see https://stackoverflow.com/questions/5717126/var-or-no-var-in-javascripts-for-in-loop

server/tools/eslint/lint-files.sh
```diff
@@ -1,4 +1,6 @@
 #! /bin/sh
 
-FILES=`find ../../app-engine/ -name "*.js"`
+# Omitting html5-qrcode as we don't want to make changes to this outside library.
+# TODO: create directory of external libraries to omit from linting
+FILES=`find ../../app-engine/ -name "*.js" ! -iname "html5-qrcode.min.js"`
```
unreasonableman, line 5: Nice!

unreasonableman (changes requested):
Two main items:
- not explicitly declaring for loop variables seems like a bad idea: https://stackoverflow.com/questions/5717126/var-or-no-var-in-javascripts-for-in-loop. There are a ton of occurrences for this. I marked the first few and then gave up. Please fix
- shouldn't we always use `let` instead of `var` to limit scope?

Also, since these are a lot of changes, please be extra paranoid about testing.

server/app-engine/static/vehicle-assignments.js
owades: good catch - simply missed this one and a few others in my find/replace

server/app-engine/static/vehicle-assignments.js
owades: updated - thanks for sharing the link about this

server/app-engine/static/vehicle-assignments.js
owades: yeah - I was trying to keep scope of this change down but...done!

server/app-engine/static/graas.js
```diff
@@ -85,15 +85,15 @@ function toRadians(degrees) {
 
 // return distance in feet between to lat/long pairs
 function getHaversineDistance(lat1, lon1, lat2, lon2) {
-    const phi1 = toRadians(lat1);
-    const phi2 = toRadians(lat2);
-    const deltaPhi = toRadians(lat2 - lat1);
-    const deltaLam = toRadians(lon2 - lon1);
+    let phi1 = toRadians(lat1);
```
unreasonableman, line 88: We should not replace `const` with `let',  just `var`. `const` has block scope already, which is what we want, plus it enforces immutability.

server/app-engine/static/graas.js
```diff
@@ -741,34 +742,33 @@ function positionCallback() {
...
+    let uuid = getUUID();
 
     if (position) {
-        lat = position.coords.latitude;
-        long = position.coords.longitude;
-        accuracy = position.coords.accuracy;
-        posTimestamp = Math.round(position.timestamp / 1000);
-        speed = position.coords.speed;
-        heading = position.coords.heading;
+        let lat = position.coords.latitude;
```
unreasonableman, line 748: I don't think this works. These variables are now scoped for the `if` block, but are later accessed outside it.

unreasonableman (changes requested):
Two sizable items (both have individual comments also):
- don't convert `const` to `let`
- there is code in `graas.js` where the scope of vars is reduced, but then those vars are accessed outside that scope

unreasonableman (approved):
Looks better now.

With the amount of changes present here, the real moment of truth is of course the testing.

owades:
Sounds good, I'm continuing to test today and hoping to deploy tonight

PR #248, opened 2022-04-26
Fix bug in pb tests

owades:
Fixed work both in staging server and local.

### DIFF
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -287,7 +287,6 @@ def add_alert(datastore_client, alert):
     print('+ wrote alert')
 
 def add_position(datastore_client, pos):
-
     entity = datastore.Entity(key=datastore_client.key('position'))
     entity.update(pos)
     datastore_client.put(entity)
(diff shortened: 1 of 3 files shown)


server/app-engine/gtfsrt.py
```diff
@@ -158,15 +159,14 @@ def purge_old_alerts(datastore_client):
...
+        print('-- alert to purge: ' + str(alert))
         key_list.append(alert.key)
 
     print('- key_list: ' + str(key_list))
 
     datastore_client.delete_multi(key_list)
     last_alert_purge = day
 
-
 ### TODO keep local cache for alert feeds
```
unreasonableman, line 170: i guess we don't need that comment anymore

unreasonableman (changes requested):
There are two remaining occurrences of `get_epoch_seconds()` in `gtfsrt.py`. They should also be changed, and ultimately the implementation of `get_epoch_seconds()` should be changed to `int(time.time())`.

unreasonableman (approved):

PR #255, opened 2022-04-28
Add animation & zoom to GRaaS Report Dashboard

owades:
Add two new features:
- Animate a trip upon click
- Zoom to a trip upon shift+click

Some general refactoring of GraphicReport as well, and retrofitting it to produce images both with and without vehicle positions.

Preview the dashboard [here](https://storage.googleapis.com/graas-resources/web/test/index.html)

One bug I just identified is that the instructions popup only appears when scrolled all the way to the top. Will work on it.

### DIFF
--- a/gtfu/src/main/java/gtfu/tools/AgencyListGenerator.java
+++ b/gtfu/src/main/java/gtfu/tools/AgencyListGenerator.java
@@ -28,7 +28,6 @@ public class AgencyListGenerator {
 
         for (int i = 0; i < objectList.size(); i++){
             String dirPath = objectList.get(i);
-            Debug.log("dirPath: " + dirPath);
             // Turns "graas-report-archive/agencyname/" into "agencyname"
             String agencyName = dirPath.substring(dirName.length(), dirPath.length() - 1);
             List <String> fileNames = gcs.getObjectList("graas-resources", dirPath);
@@ -37,16 +36,12 @@ public class AgencyListGenerator {
 
             for (int j = 0; j < fileNames.size(); j++)   {
                 String filePath = fileNames.get(j);
-                Debug.log("filePath: " + filePath);
                 // Turns "graas-report-archive/agencyname/agencyname-2022-02-07.png" into "agencyname-2022-02-07.png"
                 String fileName = filePath.substring(dirPath.length(), filePath.length());
                 if(fileName.contains(".png")) {
                     // turns "agencyname-2022-02-10.png" into "2022-02-10"
                     String date = fileName.substring(agencyName.length() + 1, fileName.length() - 4);
                     dates.add(date);
-                    Debug.log("fileName: " + fileName);
-                    Debug.log("agencyName: " + agencyName);
-                    Debug.log("date: " + date);
                 }
             }
             if(dates.size() > 0){
--- a/gtfu/src/main/java/gtfu/tools/AgencyListGenerator.java
+++ b/gtfu/src/main/java/gtfu/tools/AgencyListGenerator.java
@@ -19,15 +19,16 @@ public class AgencyListGenerator {
     public static void generateAgencyList(boolean isTest) throws IOException {
 
         GCloudStorage gcs = new GCloudStorage();
-        String dirName = "graas-report-archive/";
+        String bucketName = "graas-resources";
+        String dirName = (isTest ? "test/" : "") + "graas-report-archive/";
 
-        List <String> objectList = gcs.getObjectList("graas-resources", dirName);
+        List <String> objectList = gcs.getObjectList(bucketName, dirName);
 
         JSONObject agencies = new JSONObject();
 
         for (int i = 0; i < objectList.size(); i++){
-
             String dirPath = objectList.get(i);
+            Debug.log("dirPath: " + dirPath);
             // Turns "graas-report-archive/agencyname/" into "agencyname"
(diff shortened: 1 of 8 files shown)


gtfu/src/main/java/gtfu/tools/AgencyListGenerator.java
```diff
@@ -16,17 +16,17 @@ public class AgencyListGenerator {
...
-    public static void generateAgencyList() throws IOException {
+    public static void generateAgencyList(boolean isTest) throws IOException {
 
         GCloudStorage gcs = new GCloudStorage();
-        String dirName = "graas-report-archive/";
+        String bucketName = "graas-resources";
+        String dirName = (isTest ? "test/" : "") + "graas-report-archive/";
 
-        List <String> objectList = gcs.getObjectList("graas-resources", dirName);
+        List <String> objectList = gcs.getObjectList(bucketName, dirName);
```
unreasonableman, line 25: I like!

gtfu/src/web/graas-report-web.js
```diff
@@ -97,26 +118,57 @@ function handleAgencyChoice(){
...
-    url = `${bucketURL}/graas-report-archive/${selectedAgency}/${selectedAgency}-${selectedDate}.json`;
+    // "?nocache=" mostly forces the page to reload files each time
+    url = `${bucketURL}/graas-report-archive/${selectedAgency}/${selectedAgency}-${selectedDate}.json?nocache=${(new Date()).getTime()}`;
     loadJSON(url, processTripJSON);
 }
 
 function processTripJSON(object){
     trips.length = 0;
-    for (var i = 0; i < object.trips.length; i++) {
-        var map = {x: object["trips"][i]["boundaries"]["map-x"],
```
unreasonableman, line 108: It would be easier to read as `object.trips[i].boundaries.mapx`, which of course would require removing all dashes from field names or replacing them with `_`.

gtfu/src/web/graas-report-web.js
```diff
@@ -97,26 +118,57 @@ function handleAgencyChoice(){
...
+        // Sort points
+        points.sort(function compareFn(a, b) {
+            if (a.secs <= b.secs) {
+                return -1;
+            }
+            else return 1;
+        });
+
+        // Remove points with duplicate timestamps - this is rare but it happens.
+        // A more robust approach would be to prevent duplicate timestamps from occuring.
```
unreasonableman, line 158: I don't understand what's happening here. First time through x is 0 and y is 1. If the timestamps for the respective elements in the point list are different, increase x, making it 1 also. Then copy points[y] into points[x], which is a nop since x and y are equal. If timestamps are the same, just increase y, so now x is 0 and y is 2?
Are you sure this does what you think it does? Please provide more detailed comment.

unreasonableman (changes requested):
Looks okay overall. I don't like the isTest arg to GraphicReport, that feels kludgy, but that's not a huge deal.

But, I don't understand your point deduplication approach, see comment there.

gtfu/src/web/graas-report-web.js
```diff
@@ -182,40 +242,97 @@ function load(){
...
+                point.x *= scaleRatio;
+                point.y *= scaleRatio;
+            });
         });
         drawReport();
     };
 
     document.body.addEventListener('click', function(event) {
-        drawReport();
+        // Click event fires twice - this is a hacky way of preventing that
```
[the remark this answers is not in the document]
owades: I tried this unsuccessfully on my first pass at making this report (back then the double-event didn't have negative impact so I left it). I will give it a second look.

owades:
I don't love isTest either - the challenge I was trying to address is that it's very easy to mess up the production dashboard when testing. Perhaps rather than doing isTest I can pass the destination path as an argument, that would be cleaner.

gtfu/src/web/graas-report-web.js
owades: I do believe it does what I think. It  I should have mentioned in a comment that I modified the logic from a [blog post](https://medium.com/swlh/javascript-remove-duplicates-from-sorted-array-in-place-d3d959fb4d77). 

It basically updates the array in place so that the first N points have unique timestamps. Then I splice it it to cut off the remaining points. I will update comments to be a bit more descriptive and link the source.

gtfu/src/web/graas-report-web.js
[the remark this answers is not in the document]
owades: I made this update. Unfortunately I wasn't able to replicate the issue with or without it, so I will keep an eye on it.

gtfu/src/web/graas-report-web.js
owades: Done! Not sure why github isn't showing this as "outdated".

owades:
@unreasonableman take a look at the most recent commit, it should address all of the above.

gtfu/src/web/style.css
```diff
@@ -26,4 +31,28 @@ html, body {
...
-}
\ No newline at end of file
+}
+
+#click-here {
+  display:  none;
+}
+
+#box{
+    z-index: 2;
```
unreasonableman, line 41: And so the race to be on top begins...

gtfu/src/web/graas-report-web.js
unreasonableman: Nice!

unreasonableman (approved):
👍

PR #263, opened 2022-05-02
Features/calendar dates

owades:
Update web app to show/hide routes based on [calendar_dates.txt](https://developers.google.com/transit/gtfs/reference#calendar_datestxt).

- Update GTFU library to include Calendar Dates
- Update TripListGenerator to add dates_on and dates_off arrays to JSON file trip-names.json
- Update graas.js to include and exclude trips from dropodown menu based on contents of dates_on and dates_off

I ran tests using a test agency with the new dates_on and dates_off fields added, as well as with agencies that lack these fields in trip-names.json. Prior to deploying I will run UpdateTripNames to add the new fields for all agencies.


gtfu/src/main/java/gtfu/CalendarDateCollection.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Iterator;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+
+public class CalendarDateCollection implements Iterable<CalendarDate>, Serializable {
+    private static final String DATE_FORMAT = "yyyyMMdd";
```
unreasonableman, line 16: It doesn't seem like that's used anywhere

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -177,6 +192,26 @@ private static Map<String, String> createCustomDirectionMap() {
         return directionMap;
     }
 
+    private static Map<String, List<String>> createServiceIDMap(int exception, CalendarDateCollection calendarDates){
+        Map<String, List<String>> dateMap = new HashMap();
+        CalendarDate currentCalDate = null;
+
+        for(int i = 0; i < calendarDates.getSize(); i++){
```
unreasonableman, line 199: CalendarDateCollection implements Iterable, so you can do:
`for (CalendarDate cd : calendarDates) {
...
}`

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -177,6 +192,26 @@ private static Map<String, String> createCustomDirectionMap() {
...
+            currentCalDate = calendarDates.get(i);
+            if(currentCalDate.getExceptionType() == exception){
+                if(dateMap.get(currentCalDate.getServiceID()) == null){
```
unreasonableman, line 202: This becomes less redundant when written as:
```
Item item = cache.get(key);

if (item == null) {
    item = new Item();
    cache.put(key, item);
} 

doStuffWith(item);
```

This is a pattern that happens pretty frequently, which is why I wrote it in a more generalized way.

server/app-engine/static/gtfs-rt-util.js
```diff
@@ -173,6 +171,12 @@ if (!fetch) {
         return Math.floor(millis / 1000);
     }
 
+    exports.isNullOrUndefined = function(object) {
+        if(object === null || typeof object === 'undefined'){
```
unreasonableman, line 175: More succinctly: `return object === null || typeof object === 'undefined'`

unreasonableman (changes requested):
Please see comments

owades:
Thank you! Updated and re-requested

gtfu/src/main/java/gtfu/tools/TripListGenerator.java
```diff
@@ -177,6 +192,23 @@ private static Map<String, String> createCustomDirectionMap() {
         return directionMap;
     }
 
+    private static Map<String, List<String>> createServiceIDMap(int exception, CalendarDateCollection calendarDates){
+        Map<String, List<String>> dateMap = new HashMap();
+        for (CalendarDate cd : calendarDates){
+            if(cd.getExceptionType() == exception){
+                List<String> dateList = dateMap.get(cd.getServiceID());
```
unreasonableman, line 199: Instead of the below, it should be:
```
List<String> dateList = dateMap.get(cd.getServiceID());

if (dateList == null){
    dateList = new ArrayList();
    dateMap.put(cd.getServiceID(), dateList);
}

dateList.add(cd.getDate());
```

unreasonableman (changes requested):
Getting closer, but please see comment.

unreasonableman (approved):

PR #273, opened 2022-05-03
Add json validation to UpdateTripNames

owades:
I opted to perform validation in Java rather than as a GH action so that we can fully automate PR creation & merging without relying on GH UI

I confirmed that this feature catches:
- Trailing comma
- Missing bracket


gtfu/src/main/java/gtfu/Util.java
```diff
@@ -969,4 +971,14 @@ public boolean verify(String arg0, SSLSession arg1) {
         HttpsURLConnection.setDefaultHostnameVerifier(validHosts);
     }
 
+    public static boolean isValidJSON(String jsonString){
+        try{
+            ObjectMapper objectMapper = new ObjectMapper();
+            JsonNode jsonNode = objectMapper.readTree(jsonString);
+        } catch(JsonProcessingException e){
+            Debug.log("exception: " + e);
```
unreasonableman, line 979: We need something more descriptive here. Chances are that at some point this will show up in a log file, and then no one will know where to look for the root cause. Stack trace is always nice, and I would also include the offending json string

unreasonableman (changes requested):
Please see comment.

gtfu/src/main/java/gtfu/Util.java
owades: Thanks. I updated a bit, and here is sample output:
```* Invalid JSON detected. Error message: com.fasterxml.jackson.core.io.JsonEOFException: Unexpected end-of-input: expected close marker for Array (start marker at [Source: (StringReader); line: 1, column: 1])
 at [Source: (StringReader); line: 1074, column: 1]```

I think printing the whole list would be overkill, especially since with some agencies it is a huuuge file.

Comment (unreasonableman) [reply gtfu/src/main/java/gtfu/Util.java]: Sorry, I'm not seeing the update. Also, I meant a stack trace to show who called this routine. The details for the json are good, but we want the trace as well.

File: gtfu/src/main/java/gtfu/Util.java [hunk 2]
```diff
@@ -969,4 +971,15 @@ public boolean verify(String arg0, SSLSession arg1) {
         HttpsURLConnection.setDefaultHostnameVerifier(validHosts);
     }
 
+    public static boolean isValidJSON(String jsonString){
+        try{
+            ObjectMapper objectMapper = new ObjectMapper();
+            JsonNode jsonNode = objectMapper.readTree(jsonString);
+        } catch(JsonProcessingException e){
+            Debug.log("* Invalid JSON detected. Stack trace:");
+            e.printStackTrace();
```
unreasonableman, line 980: <pedantery>

`e.printStackTrace()` writes to stdout. `Debug.log()` currently also logs to stdout, but that may change in the future (different or additional streams).  To go the extra mile, we could use `printStackTrace(PrintStream)` with a `ByteArrayOutputStream` backed `PrintStream` and then convert that BAOS to a string and pass to `Debug.log()`.

This is not required at this point! You can just paste something along the blurb above into the code for now.

</pedantery>

unreasonableman: To get extra fancy, we could upgrade `Debug.log()` to check the type of the arg. If it's of type `Throwable` then perform the steps outlined above. That way, things would just automagically work.

unreasonableman (approved):

gtfu/src/main/java/gtfu/Util.java
unreasonableman: To continue the rambling, we can add `Debug.setStackTraceForExceptionArg(boolean)`, which would control whether or not traces are generated. The default should probably be `true`.

owades: Thank you! Added comment about this

PR #279, opened 2022-05-04
Automerge tripname changes

owades:
When small changes to trip-names.json are detected (currently defined by a % difference in byte array length of <5%), automerge the PR rather than waiting for approval.

Also add a flag to UpdateTripNames.json which runs the comparison regardless of whether or not there was a recent update. This is useful for when we update TripListGenerator output and for testing, as it prevents us from having to manually comment out code.

I tested this by editing gtfs files and ensuring that a 5% difference in byte array length correlates to what I intuitively thing that would be (ie just a few lines, depending on length of course).

### DIFF
--- a/gtfu/src/main/java/gtfu/tools/GitHubUtil.java
+++ b/gtfu/src/main/java/gtfu/tools/GitHubUtil.java
@@ -98,13 +98,11 @@ public class GitHubUtil {
     * @param file           Byte array representation of file
     * @param message        Commit message
     * @param branchName     Name for new branch
+    * @param autoMerge      Whether or not the PR should be merged automatically
     */
-    public GHPullRequest createCommitAndPR(String title, String description, String path, byte[] file, String message, String branchName) throws Exception {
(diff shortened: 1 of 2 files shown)


gtfu/src/main/java/gtfu/tools/GitHubUtil.java
```diff
@@ -85,8 +85,9 @@ private void createCommit(String branchName, String message, String path, byte[]
         GHRef branch = repo.createRef("refs/heads/" + branchName, commit.getSHA1());
     }
 
-    private void createPR(String title, String branchName, String description) throws Exception {
+    private void createPR(String title, String branchName, String description, Boolean autoMerge) throws Exception {
```
unreasonableman, line 88: As mentioned, `Boolean` is a wrapper class for the primitive `boolean` type. Unless there are very specific reasons  please use the primitive type.

unreasonableman (approved):

PR #283, opened 2022-05-09
Reduce friction for bulk assignment mode

owades:
1. When bulk assignment mode is enabled and there is a vehicle ID cached, immediately load the "posting updates" view, with no option to stop trip.
2. When bulk assignment mode is enabled and there is not yet a vehicleID cached, show the vehicle-id dropdown without a "Load trips" button. Once selected and "go" is pressed, load the "posting updates" view, with no option to stop trip.
3. Normal behavior when bulk assignment mode is not enabled. Auto populate vehicleID dropdown (it can be manually changed though)


unreasonableman (approved):
Very clean, nice!

PR #284, opened 2022-05-10
Bugfix/inconsistent training metadata

unreasonableman:
Use consistent metadata formatting


owades (approved):
Nice. I wonder if this will resolve the error I was getting in that script :-)

PR #289, opened 2022-05-13
create class for uploading live agency gtfs to bucket

owades:
Output from running ` java -cp build/libs/gtfu.jar gtfu.tools.StaticGTFSToBucket  -u <live agencies url> -b graas-resources -c src/main/resources/conf/cache` is visible [here](https://console.cloud.google.com/storage/browser/graas-resources/gtfs-archive;tab=objects?project=lat-long-prototype&prefix=&forceOnObjectsSortingFiltering=false)

### DIFF
--- a/gtfu/src/main/java/gtfu/tools/StaticGTFSToBucket.java
+++ b/gtfu/src/main/java/gtfu/tools/StaticGTFSToBucket.java
@@ -1,0 +1,154 @@
+package gtfu.tools;
+
+import gtfu.ProgressObserver;
+import gtfu.ConsoleProgressObserver;
+import gtfu.Util;
+import gtfu.Debug;
+import gtfu.EmailFailureReporter;
+import gtfu.FailureReporter;
+import gtfu.Recipients;
+
+import java.util.zip.ZipFile;
+import java.util.zip.ZipEntry;
+import java.nio.charset.StandardCharsets;
+import java.io.IOException;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.util.Enumeration;
+
+/**
+ * Uploads agency static GTFS files to GCloud Bucket
+ */
+public class StaticGTFSToBucket {
+    /**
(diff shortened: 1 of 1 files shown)


gtfu/src/main/java/gtfu/tools/StaticGTFSToBucket.java
```diff
@@ -0,0 +1,154 @@
...
+
+                int contentLength = Integer.parseInt(cl);
+                Debug.log("- contentLength: " + contentLength);
+                ConsoleProgressObserver progressObserver = new ConsoleProgressObserver(40);
+                progressObserver.setMax(contentLength);
+
+                File zf = new File(name + "/gtfs.zip");
+                FileOutputStream fos = new FileOutputStream(zf);
+
+                Util.downloadURLContent(gtfsURL, fos, progressObserver);
```
unreasonableman, line 85: You should check the modification time of the cached gtfs.zip against the modification time of the URL endpoint. If the latter isn't greater than the former, you don't need to redownload.

For extra points, you can create a version of `Util.downloadURLContent()` that takes an extra `pathToCachedVersion` arg, and then solve the general issue there instead of just the specific instance here.

unreasonableman: If the arg is `null` then don't check the modification date. Make an overloaded version of the method that has the same signature as before so you don't have to change calls to it everywhere.

unreasonableman (changes requested):
See comments.

gtfu/src/main/java/gtfu/tools/StaticGTFSToBucket.java
owades: I've made this update. I thought through what it would take to have a Util function that serves this purpose - it's a great idea since this is very similar to Util.UpdateCacheIfNeeded. I ended up deciding not to do it, because the need to check gcloud in this instance, rather than the cache, leads to a bit too much complexity.

unreasonableman (approved):

PR #293, opened 2022-05-16
trip updates clean-up

unreasonableman:
- more nuanced return values for handle_pos_update() (status and optional backfilled_trip_id)
- remove nonsense method from gtfs-rt-util module
- more flexible version of position-update test. test-position-update.js supersedes post-position-update.js


owades (approved):
Looks good, only question is do we need to keep post-position-update.js? From here it seems like it might be duplicate.

unreasonableman:
Didn't want to just yank it since it's part of pr-test. We can discuss
pr-test update steps tomorrow.

PR #298, opened 2022-05-23
add onboarding files

owades:
including test vehicle IDs.


owades:
@unreasonableman FYI, adding amtrak files so I can test triplistgenerator

PR #304, opened 2022-05-31
fix server deployment

unreasonableman:
Huffing along on the deprecation treadmill: as of today, server deploys would fail catastrophically with the server dying on startup, and clients getting `HTTP ERROR 502 Bad Gateway`.

python package and server sub-dependency `protobuf` had a release last Friday, which seems to cause serious issues. Explicitly pinning version for `protobuf` to 3.20.1.

Tested by deploying to staging, then 1) curling the target URL and 2) running a node test to push a position update. Both tests will fail if deploy is unsuccessful.


owades (approved):
good catch!

PR #305, opened 2022-05-31
Service alert UI

owades:
Service alert UI summary:
- Agency login/auth via pasting secret key
- Post a service alert. Currently this relies on the user understanding trip/route/stop IDs. I've gone back and forth on this, but fear that trying to come up with names for each of them creates too many opportunities for error (ie 4 stops have the same name, or two trips are the same except for day of week - which one do we use?)
- View service alert feed (using the actual service-alert endpoint, modified to have option of showing alerts that have future start dates)
- View and delete an individual alert

Several small updates:
- Modify updateCacheIfNeeded to not error out when optional calendar_dates.txt is missing
- Move some functions from graas.js to gtfs-rt-util.js
- update TLG to use github as source of truth rather than gcloud bucket

### DIFF
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -290,6 +290,31 @@ def add_alert(datastore_client, alert):
     datastore_client.put(entity)
     print('+ wrote alert')
 
+def delete_alert(datastore_client, alert):
+    print('delete_alert()')
+    print('- alert: ' + str(alert))
+
+    query = datastore_client.query(kind='alert')
+    # We need to filter for all fields, since a unique ID isn't passed into the protobuf
+    # query.add_filter('agency_key', '=', alert["agency_key"])
+    query.add_filter('cause', '=', alert["cause"])
+    # query.add_filter('effect', '=', alert["effect"])
+    # query.add_filter('description', '=', alert["description"])
+    query.add_filter('time_start', '=', alert["time_start"])
+    query.add_filter('time_stop', '=', alert["time_stop"])
+    # add filter for all fields
+    results = list(query.fetch(limit=20))
+    key_list = []
+
+    for alert in results:
+        print('-- alert to delete: ' + str(alert))
+        key_list.append(alert.key)
+
+    print('- key_list: ' + str(key_list))
+
+    datastore_client.delete_multi(key_list)
+    print('+ deleted alert')
+
 def add_position(datastore_client, pos):
     entity = datastore.Entity(key=datastore_client.key('position'))
     entity.update(pos)
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -59,16 +59,19 @@ def make_translated_string(text):
 def make_entity_selector(item):
     selector = gtfs_realtime_pb2.EntitySelector()
 
-    if 'agency_id' in item:
+    if 'agency_id' in item and item['agency_id'] != None:
         selector.agency_id = item['agency_id']
 
-    if 'route_id' in item:
+    if 'route_id' in item and item['route_id'] != None:
         selector.route_id = item['route_id']
 
-    if 'stop_id' in item:
+    if 'stop_id' in item and item['stop_id'] != None:
         selector.stop_id = item['stop_id']
 
-    if 'trip_id' in item:
+    if 'route_type' in item and item['route_type'] != None:
+        selector.route_type = item['route_type']
+
+    if 'trip_id' in item and item['trip_id'] != None:
         trip = gtfs_realtime_pb2.TripDescriptor()
         trip.trip_id = item['trip_id']
         selector.trip.CopyFrom(trip)
@@ -271,11 +274,12 @@ def add_alert(datastore_client, alert):
         print('alert doesn\'t have associated agency, discarding')
         return
 
-    if not('time_start' in alert and 'time_stop' in alert):
-        print('alert doesn\'t have valid time range, discarding')
-        return
+    # Per the spec, alerts without a time range will just live forever until removed
+    # if not('time_start' in alert and 'time_stop' in alert):
+    #     print('alert doesn\'t have valid time range, discarding')
+    #     return
 
-    if not('agency_id' in alert or 'route_id' in alert or 'trip_id' in alert or 'stop_id' in alert):
+    if not('agency_id' in alert or 'route_id' in alert or 'trip_id' in alert or 'stop_id' in alert  or 'route_type' in alert):
         print('alert doesn\'t have an affected entity, discarding')
         return
 
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -140,23 +140,23 @@ def alert_is_current(alert, include_future_alerts):
     if not('time_start' in alert and 'time_stop' in alert):
         return False
     now = int(time.time())
-    # No start or stop time listed. Alert lives until deleted
+    # No start or stop time listed. Is current until deleted.
     if(alert['time_stop'] == 0 and (alert['time_stop'] == 0)):
         return True
-    # Alert started in the past
+    # Alert started in the past...
     if(int(alert['time_start']) <= now):
-        # Started in the past, no stop time. Alert lives until deleted
+        # ...with no stop time. Is current until deleted.
         if(int(alert['time_stop']) == 0):
             return True
-        # Started in the past, hasn't ended
+        # ...and hasn't hasn't ended yet. Is current until stop time.
         if(int(alert['time_stop']) > now):
             return True
-        # Started and ended in the past
+        # ...and ended in the past. Is not current.
         if(int(alert['time_stop']) < now):
             return False
-    # Alert starts in the future
+    # Alert starts in the future...
     else:
-        # Include in feed if specifically requested
+        # ...is current only if include_future_alerts passed as true (for service-alert ui)
         if(include_future_alerts == 'True'):
             return True
         else:
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -309,7 +309,21 @@ def delete_alert(datastore_client, alert):
     query.add_filter('description', '=', alert["description"])
     query.add_filter('time_start', '=', alert["time_start"])
     query.add_filter('time_stop', '=', alert["time_stop"])
-    # add filter for all fields
+    if(alert["stop_id"] != ""):
+        print("filtering for stop_id...")
+        query.add_filter('stop_id', '=', alert["stop_id"])
+    if(alert["trip_id"] != None):
+        print("filtering for trip_id...")
+        query.add_filter('trip_id', '=', alert["trip_id"])
+    if(alert["route_id"] != ""):
+        print("filtering for route_id...")
+        query.add_filter('route_id', '=', alert["route_id"])
+    if(alert["agency_id"] != ""):
+        print("filtering for agency_id...")
+        query.add_filter('agency_id', '=', alert["agency_id"])
+    if(alert["route_type"] != 0):
+        print("filtering for route_type...")
+        query.add_filter('route_type', '=', alert["route_type"])
     results = list(query.fetch(limit=20))
     key_list = []
 
@@ -318,9 +332,10 @@ def delete_alert(datastore_client, alert):
         key_list.append(alert.key)
 
     print('- key_list: ' + str(key_list))
-
+    alerts_to_delete = len(key_list)
     datastore_client.delete_multi(key_list)
-    print('+ deleted alert')
+
+    print(f'+ deleted {str(alerts_to_delete)} alerts')
 
 def add_position(datastore_client, pos):
     entity = datastore.Entity(key=datastore_client.key('position'))
--- a/server/app-engine/gtfsrt.py
+++ b/server/app-engine/gtfsrt.py
@@ -140,12 +140,27 @@ def alert_is_current(alert, include_future_alerts):
     if not('time_start' in alert and 'time_stop' in alert):
         return False
     now = int(time.time())
-    if(include_future_alerts == 'True'):
-        print("include_future_alerts: " + str(include_future_alerts))
-        return now < int(alert['time_stop'])
+    # No start or stop time listed. Alert lives until deleted
+    if(alert['time_stop'] == 0 and (alert['time_stop'] == 0)):
+        return True
+    # Alert started in the past
+    if(int(alert['time_start']) <= now):
+        # Started in the past, no stop time. Alert lives until deleted
+        if(int(alert['time_stop']) == 0):
+            return True
+        # Started in the past, hasn't ended
+        if(int(alert['time_stop']) > now):
+            return True
+        # Started and ended in the past
+        if(int(alert['time_stop']) < now):
+            return False
+    # Alert starts in the future
     else:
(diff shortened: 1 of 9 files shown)


server/app-engine/gtfsrt.py
```diff
@@ -403,11 +428,7 @@ def add_alert(datastore_client, alert):
         print('alert doesn\'t have associated agency, discarding')
         return
 
-    if not('time_start' in alert and 'time_stop' in alert):
```
owades, line 406: Per the spec, alerts without a time range will just live forever until removed

owades:
@unreasonableman should I deploy to staging in order for you to test?

server/app-engine/gtfsrt.py
```diff
@@ -174,12 +177,32 @@ def make_trip_update(list):
...
 
-def alert_is_current(alert):
+def alert_is_current(alert, include_future_alerts):
     print('alert_is_current()')
     if not('time_start' in alert and 'time_stop' in alert):
         return False
     now = int(time.time())
-    return now >= int(alert['time_start']) and now < int(alert['time_stop'])
+    # No start or stop time listed. Is current until deleted.
+    if(alert['time_stop'] == 0 and (alert['time_stop'] == 0)):
```
unreasonableman, line 186: python doesn't do parens around if clauses

unreasonableman, line 186: You're checking `time_stop` twice. Also, it looks like you're using 0 to denote both the beginning of time and the end of time. May I suggest instead to use 0 for beginning of time and 2^31/1000 as the end of time? With symbolic constants of course

server/app-engine/gtfsrt.py
```diff
@@ -245,7 +270,7 @@ def get_alert_feed(datastore_client, agency):
         count = 1
 
         for item in results:
-            if alert_is_current(item):
+            if alert_is_current(item, include_future_alerts):
                 print('-- ' + str(item))
```
unreasonableman, line 274: No one looking at the logs will know what this is about.

server/app-engine/gtfsrt.py
```diff
@@ -224,7 +248,8 @@ def get_alert_feed(datastore_client, agency):
     print('- agency: ' + agency)
 
     name = agency + '-alert-feed'
-    alert_feed = cache.get(name)
+    # alert_feed = cache.get(name)
+    alert_feed = None
```
unreasonableman, line 252: What is going on here? Am I missing something or are you disabling caching of service alerts? Please explain.

server/app-engine/gtfsrt.py
```diff
@@ -418,6 +439,46 @@ def add_alert(datastore_client, alert):
...
+
+    query = datastore_client.query(kind='alert')
+    # We need to filter for all fields, since a unique ID isn't passed into the protobuf
+    query.add_filter('agency_key', '=', alert["agency_key"])
+    query.add_filter('cause', '=', alert["cause"])
+    query.add_filter('effect', '=', alert["effect"])
+    query.add_filter('description', '=', alert["description"])
+    query.add_filter('time_start', '=', alert["time_start"])
+    query.add_filter('time_stop', '=', alert["time_stop"])
+    if(alert["stop_id"] != ""):
```
unreasonableman, line 454: Without having seen all of your changes, I will assume that delete_alerts is a public interface. How do you know that clients will set unused fields to the empty string, vs. `None` or something else?

server/app-engine/gtfsrt.py
```diff
@@ -418,6 +439,46 @@ def add_alert(datastore_client, alert):
...
+
+    for alert in results:
+        print('-- alert to delete: ' + str(alert))
+        key_list.append(alert.key)
+
+    print('- key_list: ' + str(key_list))
+    alerts_to_delete = len(key_list)
+    datastore_client.delete_multi(key_list)
+
+    print(f'+ deleted {str(alerts_to_delete)} alerts')
```
unreasonableman, line 480: `f` and `{}` is smart enough to to do `str()` on expressions if needed

server/app-engine/main.py
```diff
@@ -104,15 +104,19 @@
...
 def service_alerts():
     print('/service-alerts.pb')
+
     agency = request.args.get('agency')
     print('- agency: ' + agency)
-
     if agency is None:
         return 'No agency given', 400
 
+    include_future_alerts = request.args.get("include_future_alerts") or "False"
```
unreasonableman, line 113: The concept of including or not including future alerts seems like a design decision that should be discussed before going to implementation. Again, I haven't see all of the changes yet.

server/app-engine/main.py
```diff
@@ -231,6 +245,32 @@ def post_alert():
 
     return Response('{"command": "post-alert", "status": "ok"}', mimetype='application/json')
 
+@app.route('/delete-alert', methods=['POST'])
+def delete_alert():
+    print('/post-alert')
```
unreasonableman, line 250: copy/paste error

server/app-engine/static/service-alert-ui.js
```diff
@@ -0,0 +1,612 @@
...
+  const baseURL = `https://storage.googleapis.com/graas-resources/gtfs-archive/${agencyID}/`;
+
+  for (let i = 0; i < files.length; i++) {
+    let fileName = files[i];
+    let url = baseURL + fileName + ".txt";
+    util.log('- fetching from ' + url);
+    util.timedFetch(url, {
+        method: 'GET'
+    })
+    .then(function(response) {
```
unreasonableman, line 97: We should use async/await for all new code if at all possible

server/app-engine/static/service-alert-ui.js
```diff
@@ -0,0 +1,612 @@
...
+    };
+
+    if(url !== null){
+      urlData = {url: url}
+      Object.assign(data, urlData)
+    }
+
+    util.signAndPost(data, signatureKey, '/post-alert', document);
+
+    // Consider actually confirming send status
```
unreasonableman, line 438: Are you saying it's unclear if post succeeded? If so, then we shouldn't proclaim success.

unreasonableman (changes requested):
See comments.

unreasonableman:
Sure

server/app-engine/gtfsrt.py
owades: The HTML datetime selector sets the value to zero when no time is selected - I opted to leave as is rather than setting to null (which would be more descriptive since 0 is technically 1970). Another option would be to set them to 0 and 2^31 in javascript, as you said, and then have simpler logic here in python. I kind of like that.

Re: time_stop twice, drat! good catch.

server/app-engine/main.py
owades: Happy to discuss (here are in meeting tomorrow), but I do think the use cases are pretty straightforward:
- For generating the protobuf file for wayfinding apps, it makes sense to only show alerts that are in effect right now.
- For the purpose of managing service alerts, it would be misleading to omit alerts that aren't in effect yet, so I made the call to show future ones as well.

server/app-engine/main.py
owades: drat!

server/app-engine/static/service-alert-ui.js
owades: Yes, it's not good, I need to update this

server/app-engine/gtfsrt.py
owades: I'm not sure what you mean by the logs, but I'm adding a note to service_alerts() in main.py to explain what this field is for. Does that address what you're referring to?

unreasonableman:
I mean if the log has lines like `-- <stuff>` then no one looking at the
logs later will know what that means, unless they dig through the source,
and maybe not even then. If we log items like these, there should be a
header statement like `- current alerts:`

owades:
Ah I see what you mean now. thanks.

owades:
@unreasonableman you can view the ui [here](https://20220603t152615-dot-lat-long-prototype.wl.r.appspot.com/service-alert-ui) and sign in using test agency creds.

unreasonableman:
Disabling the cache could have major performance and cost consequences.
It's all theoretical now with the couple of agencies that we have, but if
we ever get to higher traffic volumes caching will make a significant
difference.

Rather than disabling caching for service alerts (in general or
selectively) I would suggest the following: put up a notice that says
something like "waiting for updates to post to system" and then hit the
service alerts endpoint every 15 seconds until the new alerts are present
in the feed. Feeds are cached for 60 seconds, so you're guaranteed to not
have to wait longer than that. In the notice, say that it will take 60
seconds or less, and every time you hit the endpoint, deduct 15 seconds
from the total, so users see that something is happening.

On a general note, something like disabling caches needs to be marked up
with giant alarm bells for commits or PRs.

owades:
Several updates:
- Create an additional service alert pb endpoint that requires auth. This one doesn't use cache, and the public one still does.
- When stop_time is null, set it to Number.MAX_VALUE, so that we don't need an extra line of python to handle null/zero dates.
- Generalize util.apiCall and util.signAndPost() to return response rather than response.json(), so they can be used more universally.
- Create python util.is_null_or_empty

owades:
New bug - delete functionality not working properly.

server/app-engine/gtfsrt.py
```diff
@@ -176,12 +176,30 @@ def make_trip_update(list):
 
     return entity
 
-def alert_is_current(alert):
+def alert_is_current(alert, service_alert_ui):
```
unreasonableman, line 179: `service_alert_ui` is a pretty terrible name for this arg. I feel like it was called `include_future_alerts` before, and that's a much better name. There is no need for the server to know anything about a service alert UI.

server/app-engine/gtfsrt.py
```diff
@@ -176,12 +176,30 @@ def make_trip_update(list):
...
     print('alert_is_current()')
     if not('time_start' in alert and 'time_stop' in alert):
         return False
     now = int(time.time())
-    return now >= int(alert['time_start']) and now < int(alert['time_stop'])
+
```
unreasonableman, line 184: Shouldn't all of these time values already be ints? Why do they keep getting cast to int?

Also, again, unless I'm missing something, this whole portion could be simplified quite a bit. If alerts come in with `time_stop` set to 0, set it to some very high value. Then, instead of the cascaded `if`s:

```
return (now >= int(alert['time_start']) or include_future_alerts) and now < int(alert['time_stop'])
```

server/app-engine/gtfsrt.py
```diff
@@ -211,12 +229,13 @@ def purge_old_alerts(datastore_client):
     datastore_client.delete_multi(key_list)
     last_alert_purge = day
 
-def get_alert_feed(datastore_client, agency):
+def get_alert_feed(datastore_client, agency, service_alert_ui):
```
unreasonableman, line 232: Same deal about the name, `use_cache` is more descriptive and doesn't introduce the concept of a service alert UI to the server.

server/app-engine/gtfsrt.py
```diff
@@ -226,7 +245,10 @@ def get_alert_feed(datastore_client, agency):
     print('- agency: ' + agency)
 
     name = agency + '-alert-feed'
-    alert_feed = cache.get(name)
+    if not service_alert_ui:
```
unreasonableman, line 248: python has a trigraph construct, so you can simplify to: `alert_feed = cache.get(name) if use_cache else None`

server/app-engine/gtfsrt.py
```diff
@@ -420,6 +438,41 @@ def add_alert(datastore_client, alert):
     datastore_client.put(entity)
     print('+ wrote alert')
 
+def delete_alert(datastore_client, alert):
+    print('delete_alert()')
+    print('- alert: ' + str(alert))
+
+    query = datastore_client.query(kind='alert')
+    # We need to filter for all fields, since a unique ID isn't passed into the protobuf
+    query.add_filter('agency_key', '=', alert["agency_key"])
```
unreasonableman, line 447: Have you tested this? I'm pretty sure that this filter group would require an update to index.yaml, which I'm not seeing as part of this PR.

server/app-engine/gtfsrt.py
owades: The other option would be to have two new boolean parameters for get_alert_feed():
1. use_cache
2. include_future_alerts
I had consolidated them into "service_alert_ui", and I see what you mean about that not being relevant to the API. I'll update.

server/app-engine/static/vehicle-assignments.js
```diff
@@ -87,24 +87,6 @@ function showToast(text) {
     }, 2000);
 }
 
-function handleModal(name) {
```
unreasonableman, line 90: It looks like you made actual code changes to the dispatch UI. Please be sure you've done enough testing for these changes.

server/app-engine/static/service-alert-ui.js
```diff
@@ -0,0 +1,613 @@
+
```
unreasonableman, line 1: This is too much new code for me to review in-depth in a reasonable amount of time.

unreasonableman (changes requested):
Please see comments. We can have a meeting around this and go over the items if you like.

server/app-engine/main.py
```diff
@@ -54,6 +62,9 @@
 
     /service-alerts.pb?agency=MY-AGENCY-ID
 
+Authenticated agencies can retrieve current service alerts, without caching, through the following endpoint:
+    /service-alert-ui.pb
```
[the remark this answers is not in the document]
owades: The issue is that in order to authenticate, I believe I need to make a POST request including agencyID and signature. I can't require that info for the existing /service-alert.pb so I needed to make a new endpoint that handles this additional data.

server/app-engine/gtfsrt.py
owades: You're right, I think it's less pretty but it does save a bunch of lines

unreasonableman:
Fair enough.

server/app-engine/gtfsrt.py
owades: thats very nice, thanks

server/app-engine/gtfsrt.py
owades: It does work - I've never modified index.yaml and don't know a ton about it. I've tested for all of the field values, though, and it successfully deletes the correct alert.

unreasonableman:
That is bizarre.

Every time I've created new filter queries I had to not just update the
index file but also deploy it, which takes a few minutes to complete.
Otherwise the server will complain about missing indices and not run the
queries.

owades:
Latest version available [here](https://20220609t133250-dot-lat-long-prototype.wl.r.appspot.com/service-alert-ui)

server/app-engine/gtfsrt.py
owades: Ok, a bit more information about indexes, from [here](https://cloud.google.com/datastore/docs/concepts/indexes) (See "Index Configuration"):

**Built-in indexes are sufficient to perform many simple queries, such as equality-only queries and simple inequality queries.**

unreasonableman:
Mystery solved! 🚀

unreasonableman (approved):
Reviewed the following commits from today and yesterday:

e3aae13
97d7b2f
979bf8f

unreasonableman:
Ugly, but a simple test succeeded, see screenshot.

PR #307, opened 2022-06-02
Examine stress test and DB performance

unreasonableman:
- batch DB writes for `current-position`, `position` and `stop-time`. These client updates will come in regularly while a trip is in progress. The way batching is implemented means that the last update for either entity may or may not be written to the DB in a timely fashion. This shouldn't be a an issue except for tests that will fail if specific updates aren't found in the DB
- add 'flush' updates to these tests to avoid the issue described above


owades:
Did you mean to include the 3 node test files in this?

owades (approved):
clever solution for batch writing. can you confirm that it will flush once the oldest update in the queue is a second old?

unreasonableman:
That's not really how it works. When a new entry is added to the batch
writer, it checks whether it's been more than the max batch time (1 second
for us here). If not, add entry to queue (which is actually a dictionary
using DB keys as keys so that we don't do concurrent updates for the same
entity), otherwise write queue entries to DB and reset timer.

As I think I mentioned in a couple of places, there is no guarantee when
the last set of entries in the queue will be written, once all updates
stop. Any post of a position, current-position or stop-time entity can
trigger a flush, but if there aren't any of those at all for the entire
server then stuff will sit in the queue until the next update rolls in.
Again, can be any update at all, doesn't have to be for the same vehicle,
trip or agency. Service alerts aren't batched as they are unlikely to be
high volume.

I don't think this is an issue, but interested to hear if you think
otherwise.

unreasonableman:
Yes, updated to work around the batch issue.

PR #311, opened 2022-06-09
bugfix for block data generator

unreasonableman:
When we create the validity period for a calendar, make it so that it spans until the end of the last day, not just the beginning of the last day.

This caused problems for the block data generator for sure, but probably also led to other failures


owades (approved):
good catch

PR #318, opened 2022-06-21
Bugfix/time zone

unreasonableman:
Let server actually use PT for time calculations.


unreasonableman:
Tested against staging:
- run web app with test agency and vehicle 4001 and observe updates in server log
- run dispatch UI and assign a block for 4001 for today
- observer backfilling of trip ID from block assignment in server logs
- run post-vehicle-positions.js node test successfully
- run post-service-alerts.js node test successfully
- run post-stop-time-entities.js node test successfully

unreasonableman:
Yes, I removed stuff that isn't called from anywhere.

owades (approved):

PR #322, opened 2022-06-30
update setup docs

unreasonableman:
- autonumbering for lists in new-instance-setup.md
- add command to check that billing is enabled
- add missing path to cloudless run instructions
- optional cloudless step to put server online


unreasonableman:
Fix was way harder than it should have been >:[

owades (approved):
beautifully done

PR #323, opened 2022-07-01
Trip inference initial port

owades:
Lightly tested trip inference + web app code, mostly in case you want to get a head-start on reviewing. Two big warnings:
- This produces a tripID, but I have not yet tested that it's the correct tripID
- Web app does not yet handle exceptions, such as what happens when a null trip ID is assigned.
- Currently web app asks driver to enter vehicleID. I'm curious what you think we should do.


server/app-engine/static/area.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    update(lat, lon){
+        // util.log("update()");
+        this.topLeft.setLonIfLess(lon)
+        this.topLeft.setLatIfGreater(lat)
+
+        this.bottomRight.setLonIfGreater(lon)
+        this.bottomRight.setLatIfLess(lat)
+        if(lat.length > 20){
```
unreasonableman, line 28: What's this all about?

server/app-engine/static/shapepoint.js
```diff
@@ -1,7 +1,48 @@
...
+        if(lat !== null){
+            this.lat = parseFloat(lat);
+        } else this.lat = lat;
+        if(lon !== null){
+            this.lon = parseFloat(lon);
+        } else this.lon = lon;
+    }
+
+    setLatIfLess(lat){
+        let latFloat = parseFloat(lat);
```
unreasonableman, line 13: The args passed into these functions appear to be strings, whereas in the python version they are numerical. Port should work the same as original wherever possible

server/app-engine/static/graas.js
```diff
@@ -921,6 +922,9 @@ function agencyIDCallback(response) {
         let arg = Date.now()
         util.log("- arg: " + arg);
         getURLContent(agencyID, arg);
+        // need to decide how to determine vehicle id
+        inf = await new TripInference('~/tmp/',agencyID, vehicleID, 15);
```
unreasonableman, line 926: The constructor isn't `async` and just assigns a few instance fields, why is there an `await`?

server/app-engine/static/gtfs-rt-util.js
```diff
@@ -353,6 +380,90 @@ if (!fetch) {
...
+        let minutes = parseInt(seconds / 60);
+        seconds -= minutes * 60;
+        return `${hours}:${this.padIfShort(minutes)}`;
+    }
+
+    exports.degreesToRadians = function(degrees){
+        return degrees * (Math.PI/180);
+    }
+
+    exports.haversineDistance = async function(lat1, lon1, lat2, lon2){
```
unreasonableman, line 418: Why is this `async`?

server/app-engine/static/graas.js
```diff
@@ -1017,7 +1025,7 @@ function loadTrips() {
             // util.log(`- lon: ${lon}`);
             const timeDelta = util.getTimeDelta(time);
             // util.log(`- timeDelta: ${timeDelta}`);
-            const distance = getHaversineDistance(lat, lon, startLat, startLon);
+            const distance = await util.haversineDistance(lat, lon, startLat, startLon);
```
unreasonableman, line 1028: `haversineDistance()` is just a few calculations, why the `await`?

server/app-engine/static/area.js
owades: leftover debugging, oops!

server/app-engine/static/inference.js
```diff
@@ -0,0 +1,830 @@
...
+        };
+
+        if (segmentList === null){
+            return ret;
+        }
+
+        // util.log(`- segmentList.length: ${segmentList.length}`);
+        // util.log(`- tripIdFromBlock: ${tripIdFromBlock}`);
+
+        let stop_id = await this.getStopForPosition(lat, lon, STOP_PROXIMITY);
```
unreasonableman, line 666: `await` not necessary from what I can see.

server/app-engine/static/segment.js
```diff
@@ -0,0 +1,120 @@
...
+
+    setSegmentsPerTrip(count){
+        this.segments_per_trip = count;
+    }
+
+    getTripFraction(){
+        return this.segment_index / this.segments_per_trip;
+    }
+
+    async getScore(lat, lon, seconds, shapes){
```
unreasonableman, line 43: don't think this needs `async`

unreasonableman (changes requested):
No hair-raising issues on first glance, but please fix gratuitous await/async and passing floats as strings

unreasonableman:
You probably want to understand why that is (e.g. because
haversineDistance() is declared as async).

PR #357, opened 2022-10-10
run trip inference in the browser, browser harness for trip inference test suite

unreasonableman:
see [here](https://github.com/cal-itp/graas/blob/features/trip-inference-js/docs/testing/trip-inference.md) for usage.


owades (approved):
I looked through much of the code, putting more energy into graas.js. Looks good, especially since you reported the results are matching and I assume there's a level of browser UX you performed.

Left a few comments about small items.

unreasonableman:
@owades requesting re-review. As mentioned over slack, I did a fair amount of testing for this latest commit, both locally and on staging. You only need to review changes from the latest commit.

owades (approved):
looks good! sorry for the delay, I'm locked out of my DOT email at the moment

PR #358, opened 2022-10-12
Add local version of .github/workflows/pr-test.yaml.

unreasonableman:
Being able to only debug test failures through git push is very cumbersome. An added wrinkle here is that it appears as if github always runs the main branch version of pr-test.yaml, which in a way is of course what you want, but also makes debugging test issues in a branch that much hairier.

Run local smoke test with tools/smoketest.sh.

Test will bail if the following env vars aren't set:

  PR_TEST_URL: server URL
  PR_TEST_ID_ECDSA: private key for pr-test agency
  PR_TEST_GRAAS_VENV: path to python virtual env created from server/app-engine/requirements.txt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ server/app-engine/db-sim.json*
 server/app-engine/ad-hoc
 server/app-engine/ad-hoc.pub
 server/app-engine/ad-hoc.png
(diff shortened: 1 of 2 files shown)


owades (approved):
thanks for creating this!

PR #373, opened 2022-11-29
save raw zip file to grass-resources/gtfs-mirror/agency-name.zip

owades:
I updated the existing gtfs archive process to save a copy of the raw zip file to 
 [graas-resources/gtfs-mirror](https://console.cloud.google.com/storage/browser/graas-resources/gtfs-mirror;tab=objects?project=lat-long-prototype&prefix=&forceOnObjectsSortingFiltering=false). Please look at the directory and let me know if this matches the need!


unreasonableman (approved):
👍

PR #391, opened 2023-01-12
- instead of adding a click listener to the entire document body, add…

unreasonableman:
- instead of adding a click listener to the entire document body, add only to timeline and map canvas
- previous download mechanism doesn't really work anymore, since routes are now drawn dynamically on top of static images
- implement functiong download as
  + create offscreen canvas
  + draw contents of timeline and maps canvas to offscreen canvas
  + convert canvas contents to data URL
  + trigger download of data URL to client machine

### DIFF
--- a/gtfu/src/web/graas-report-web.js
+++ b/gtfu/src/web/graas-report-web.js
@@ -64,6 +64,10 @@ const tooltipItems = ["trip_id", "vehicle_id", "agent", "device",
  const ALPHA_MIN = 0.4;
  const MARGIN = 5;
 
+function log(s) {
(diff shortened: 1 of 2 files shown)


unreasonableman:
@owades What's the deployment route for the graas report web app? Just copy relevant files to the bucket, or something else?

owades:
> @owades What's the deployment route for the graas report web app? Just copy relevant files to the bucket, or something else?

That's right - copy relevant files to the bucket.

owades (approved):
This looks and works great! I pressure tested locally, clicking everywhere and trying all the features.

PR #405, opened 2023-04-10
Be more thoughtful about removing agency token. 

unreasonableman:
We previously removed for any number of reasons, including no network connectivity. Adapted to only remove upon failure to load key from token, which is a good indicator that token is corrupt.

Tested by temporarily changing the `/hello` URL to something bogus  and back repeatedly, and observing resulting behavior.


owades (approved):
thank you!

PR #407, opened 2023-04-20
automation groundwork

unreasonableman:
- create grass bot that collects usage stats and posts to slack channel at 11 pm PT every day
- make all server endpoints follow the designated signature verification approach, instead of fragmented local variations
- fix issues around initial token verification 
- allow python code to produce grass-compatible ECDSA keys with a couple of lines of code, instead of tedious command line sequences or brittle system calls from within python (see ecdsa-test.py for an example on how to generate)
- tested token verification, position updates and alert updates

### DIFF
--- a/.github/workflows/graas-bot.yaml
+++ b/.github/workflows/graas-bot.yaml
@@ -1,0 +1,29 @@
+name: Collect daily GRaaS bot stats
+on:
+  # Daily at 11PM pacific
+  schedule:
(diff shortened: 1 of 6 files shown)


owades (approved):
looks good! that table seemed fun to make.

unreasonableman:
> looks good! that table seemed fun to make.

As you know, I'm all business all the time :|

PR #408, opened 2023-04-24
feed expiration warning

owades:
New feature that alerts graas_internal Slack channel when a GRaaS agency's schedule feed is expiring within 2 weeks (or has already expired).

CheckFeedExpiration can also be set to send email alerts (or both email and Slack) via command line.

New classes:
- FeedInfo
- CheckFeedExpiration
- Slack
- SlackFailureReporter

This PR also updates UpdateTripNames to only send an email when a PR is created.


gtfu/src/main/java/gtfu/Time.java
```diff
@@ -197,4 +198,19 @@ public static String formatDate(String format, Date date) {
...
+    public static Date getDateFromYYYYMMDD(String yyyyMMDD) {
+        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyyMMdd");
+        try {
+            return dateFormat.parse(yyyyMMDD);
+        } catch (ParseException e) {
+            throw new RuntimeException("Error parsing date: " + e.getMessage(), e);
+        }
+    }
+
+    public static int getDaysUntilDateLong(String yyyyMMDD) {
```
unreasonableman, line 211: It maybe more versatile to have something like `getDayDifference(String yyyyMMDD)`, which doesn't make assumptions about whether the date passed in is before or after the current date. For extra points you can pass two dates into the method, and then have a convenience method that only takes one arg, and sets the second one to `today`. Method returns negative number if arg 1 is less than arg2, i.e. target date is before reference date, positive number otherwise.

gtfu/src/main/java/gtfu/tools/CheckFeedExpiration.java
```diff
@@ -0,0 +1,144 @@
...
+ * Daily checks and updates on each active agency's feed.
+ */
+public class CheckFeedExpiration {
+    private static final long MAX_FEED_EXPIRATION_WINDOW_DAYS = 14;
+
+     /**
+     * Runs CheckFeedExpiration for a single agency
+     * @param agencyID The agencyID
+     */
+    public static void CheckFeedExpiration(String agencyID, boolean sendEmailAlert, boolean sendSlackAlert) throws Exception {
```
unreasonableman, line 31: methods should start with a lowercase character (more than one occurrence in this file.

gtfu/src/main/java/gtfu/tools/Slack.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class Slack {
+    private String webHookURL;
+
+    /**
+     * Initialize Slack connection
+     */
+    public Slack() {
+        this.webHookURL = System.getenv("GRAAS_BOT_WEBHOOK");
+        if(webHookURL == null){
```
unreasonableman, line 17: If `webhookURL` is null, then it's over. Nothing will work, so we should do more than just do a `Debug.log()` Remember that debug logging may be turned off in production. We should at least do `Debug.error()` and throw an exception.

gtfu/src/main/java/gtfu/tools/UpdateTripNames.java
```diff
@@ -117,10 +117,10 @@ public static void UpdateTripNames(String[] agencyIDList, boolean regenerateAll)
                 Debug.log("trip-names.json has been updated since the last static GTFS update. Continuing.");
             }
         }
-        if (prCount == 0) {
-            reporter.addLine("...no agencies. Everything looks up to date.");
+        // Send email only if a PR has been created.
+        if (prCount > 0) {
```
unreasonableman, line 121: Improved signal to noise ratio. I like!

unreasonableman (changes requested):
The `Time` thing is kind of optional to fix, please take care of the other two though.

gtfu/src/main/java/gtfu/Time.java
owades: updated! I ended up doing another "getDaysUntilDate" function, but one that utilizes a "differenceBetweenDates" function. The resulting logic is a little funny, as types get converted back and forth and back, but I think result is something that is intuitive to interact with.

gtfu/src/main/java/gtfu/tools/CheckFeedExpiration.java
owades: done thanks!

gtfu/src/main/java/gtfu/tools/Slack.java
owades: thanks, done!

gtfu/src/main/java/gtfu/Time.java
```diff
@@ -197,4 +197,33 @@ public static String formatDate(String format, Date date) {
...
+    }
+
+    public static int getDayDifferenceInt(String date1YYYYMMDD, String date2YYYYMMDD) {
+        Date date1 = getDateFromYYYYMMDD(date1YYYYMMDD);
+        Date date2 = getDateFromYYYYMMDD(date2YYYYMMDD);
+
+        long diffInMillis = date1.getTime() - date2.getTime();
+        long daysUntilEnd = diffInMillis / MILLIS_PER_DAY;
+
+        return (int) daysUntilEnd;
```
unreasonableman, line 222: Complete nitpick: this will cause an under/overflow for dates more than 2B days apart. However that's more than 5M years, and so not a practical concern!

gtfu/src/main/java/gtfu/tools/CheckFeedExpiration.java
```diff
@@ -0,0 +1,137 @@
...
+ * Daily checks and updates on each active agency's feed.
+ */
+public class CheckFeedExpiration {
+    private static final long MAX_FEED_EXPIRATION_WINDOW_DAYS = 14;
+
+     /**
+     * Runs CheckFeedExpiration for a single agency
+     * @param agencyID The agencyID
+     */
+    public static void checkFeedExpiration(String agencyID, boolean sendEmailAlert, boolean sendSlackAlert) throws Exception {
```
unreasonableman, line 26: A little better, but still not good. Naming a method of a class the same as that class is confusing, casing aside. I recommend just doing a regular constructor and then creating an instance from main():

```
CheckFeedExpiration cfe = new CheckFeedExpiration(/*args*/);
```

 This also allows intuitive use of this class from other code.

gtfu/src/main/java/gtfu/tools/Slack.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class Slack {
+    private String webHookURL;
+
+    /**
+     * Initialize Slack connection
+     */
+    public Slack() {
+        this.webHookURL = System.getenv("GRAAS_BOT_WEBHOOK");
+        Debug.error("* No env variable found with name GRAAS_BOT_WEBHOOK...exiting.");
```
unreasonableman, line 17: Something happened here. The null check of webHookURL disappeared. Also a reminder to not call System.exit() from code that may be called from other code, as it may lead to very unpleasant debugging down the road.

unreasonableman (changes requested):
See inline comments.

gtfu/src/main/java/gtfu/tools/Slack.java
owades: oof!!

owades:
thank you for these. clearly a bit rusty over here...

unreasonableman (approved):

PR #420, opened 2023-08-17
add support for realtime feeds that require an API key for access

unreasonableman:
reminder: use `./scripts/monitor-position-updates.sh` to run MonitorPositionUpdates tool, `-url <url>` to set feed URL and `-api-key <api-key>` to set API key.

Tested with NYC MTA.


owades (approved):
Nice!

